PR #3, opened 2016-04-05
Pipeline support for C#

chrisdunelm:


tbetbetbe:
LGTM

I'm puzzled as to why the travis CI run was not triggered
Please submit, in the event of any issues, we can fix it in a follow-up

PR #4, opened 2016-04-05
Config generation pipeline

michaelbausor:
Added config generation pipeline and associated tests


pipeline/pipelines/code_generation_pipeline.py
```diff
@@ -271,3 +271,20 @@ def do_build_flow(self, **kwargs):
 
     def validate_kwargs(self, **kwargs):
         _validate_codegen_kwargs(_VGEN_REQUIRED, **kwargs)
+
+
+class VkitConfigPipeline(pipeline_base.PipelineBase):
```
garrettjonesgoogle: I would suggest putting this before the other pipelines, because in the process of building a client, this would be the first thing that is used.

michaelbausor: Moved to top

garrettjonesgoogle:
Two more things to do (after observing it run over the shoulder):

* update the commands in USAGE.rst to refer to artman
* write to ${api_name}_veneer.yaml (as specified in the pipeline config, which will be a prerequisite for this pipeline)

garrettjonesgoogle:
Oh, and also add a sample command of configgen to USAGE.rst .

michaelbausor:
Still remaining to update USAGE.rst to refer to artman paths - needs to happen at same time as update to googleapis repo

test/testdata/config_pipeline.baseline
```diff
@@ -0,0 +1,4 @@
+mkdir -p {OUTPUT}
+protoc --proto_path=test/fake-repos/gapi-core-proto/src/main/proto/ --proto_path=test/fake-repos/fake-proto --proto_path=MOCK_GRADLE_TASK_OUTPUT --include_imports --include_source_info -o {OUTPUT}/library-v1.desc test/fake-repos/fake-proto/fake.proto
+mkdir -p {OUTPUT}/library-v1-config-gen
+test/fake-repos/gapi-tools/gradlew -p test/fake-repos/gapi-tools runConfigGen -Pclargs=--descriptor_set={OUTPUT}/library-v1.desc,--output={OUTPUT}/library-v1-config-gen/library-v1_veneer.yaml
```
garrettjonesgoogle: Does this baseline file need to be updated for your latest change? (doing the mv command)

pipeline/tasks/veneer_tasks.py
```diff
@@ -23,6 +23,50 @@
...
+
+    def validate(self):
+        return [vgen_requirements.ConfigGenRequirements]
+
+
+class VeneerConfigMoveTask(task_base.TaskBase):
+    """Move config file to veneer_api_yaml location"""
+
+    def execute(self, intermediate_config_location, veneer_api_yaml):
+        for conf_out in veneer_api_yaml:
```
garrettjonesgoogle: This has the odd behavior of writing the same file to multiple output files if someone decides to specify multiple. Maybe we should error out if multiple are provided. The idea would be that you can only generate into a single file, but later, you can split it up and configure multiple files to feed into codegen.

test/testdata/config_pipeline.baseline
michaelbausor: Updated test baseline and added missing calls

pipeline/tasks/veneer_tasks.py
michaelbausor: Now checks for multiple outputs. Currently it just prints an error message, but does not fail the process (i.e. by throwing an exception). Is this what we want?

garrettjonesgoogle: I think we should fail the process - we can't produce what the user is expecting in this case.

michaelbausor: Updated

pipeline/tasks/veneer_tasks.py
```diff
@@ -55,10 +55,10 @@ def _move_to(self, intermediate_config_location, veneer_api_yaml):
                         os.path.abspath(intermediate_config_location),
                         [os.path.abspath(c_out) for c_out in veneer_api_yaml])
 
-        if len(veneer_api_yaml) == 0:
+        if len(veneer_api_yaml) > 1:
+            raise ValueError(error_fmt + 'Multiple locations specified')
+        elif len(veneer_api_yaml) == 0:
             print error_fmt + 'No location specified'
```
garrettjonesgoogle: The rest of these cases should probably cause it to fail also.

michaelbausor: Updated. Note that this will cause consecutive runs of the config generator to fail, because the file will already exist. Will this be OK? I thought yes because we don't expect the config generator to be run repeatedly, just once and then make manual edits to the file.

garrettjonesgoogle: I think that should be ok. If we decide to add 3-way merge support later, then we can avoid that kind of failure.

michaelbausor:
Are there any outstanding issues on this PR? There is no need for extra changes to support artman in usage docs, as they reference pipeline folder in googleapis not the pipeline repo

garrettjonesgoogle:
LGTM

PR #5, opened 2016-04-07
Java packman

michaelbausor:
Updated packman task to support multiple languages
Updated java gprc task to use packman
Updated tests


pipeline/pipelines/code_generation_pipeline.py
```diff
@@ -74,11 +74,10 @@ def validate_kwargs(self, **kwargs):
         _validate_codegen_kwargs([], **kwargs)
 
 
-class PythonGrpcClientPipeline(pipeline_base.PipelineBase):
+class GrpcClientPipelineBase(pipeline_base.PipelineBase):
```
garrettjonesgoogle: We have been keeping the different language pipelines separate from each other so that they can be changed without disturbing the other languages. I think we might want to keep it that way for the moment.

michaelbausor: Removed dependency on each other

PR #10, opened 2016-04-21
Consistently use GAPIC as the project name

bjwatson:
Pre-push hook installed.

Change-Id: I6599f5d570f0c36d56db48ce599a40babe2b9479


pipeline/pipelines/pipeline_factory.py
```diff
@@ -22,9 +22,9 @@
 
 
 def make_pipeline_flow(pipeline_name, **kwargs):
-    """This factory function to make veneer pipeline.
+    """Factory function to make a GAPIC pipeline.
 
-    Because veneer pipeline is using OpenStack Taskflow, this factory function
+    Because GAPIC pipeline is using OpenStack Taskflow, this factory function
```
garrettjonesgoogle: -> "Because the GAPIC pipeline"

tbetbetbe:
N.B.  There is a lint error causing the CI to fail.
It may not be related to this change, but given that this is cleanup CL, I think it should be fixed here.
Going forward, we should refuse to approving any CL unless travis is green.

pipeline/pipelines/pipeline_factory.py
bjwatson: Good catch. Annoying to fix since I had to reformat all the lines (pep8 strictly enforces an 80 character limit per line).

bjwatson:
@tbetbetbe Agreed about keeping Travis results clean. This was caused by this change because lint is strict about indentation. I changed the length of several first lines with veneer -> gapic, which threw off the indentation of the second line. Slightly annoying facet of Python style... :/

garrettjonesgoogle:
LGTM

PR #16, opened 2016-05-02
Update Dockerfile and requirements.txt

ethanbao:
Modify Dockerfile and requirements.txt to include more dependency installation in order to support running pipeline jobs remotely.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,37 @@
 FROM ubuntu:14.04
 
+# Set the locale
+RUN locale-gen en_US.UTF-8
+ENV LANG en_US.UTF-8
+ENV LANGUAGE en_US:en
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,15 +1,35 @@
...
+ENV LANGUAGE en_US:en
+ENV LC_ALL en_US.UTF-8
+
 # Install essential packages.
 RUN apt-get update && apt-get install -y --no-install-recommends \
     build-essential \
     man-db \
     pkg-config \
+    libffi-dev \
+    libssl-dev \
```
jmuk: Who requires these two?

ethanbao: gcloud-python needs these two. I added comments.

jmuk:
cool, I think it's good to go -- LGTM.

PR #19, opened 2016-05-06
Update USAGE for trimming API name prefix

jmuk:
Related to https://github.com/googleapis/googleapis/pull/22.
If this PR happens, USAGE should be updated like this.


garrettjonesgoogle:
Nice improvement. LGTM.

PR #22, opened 2016-05-11
Introduce a new pipeline for remote execution

jmuk:
The new pipeline loads the pipeline config files as a step of
tasks and passes the loaded arguments to the actual pipeline.

This helps for remote execution, so the users don't have to
set up the configs or protos repository locally.

### DIFF
--- a/pipeline/pipelines/pipeline_factory.py
+++ b/pipeline/pipelines/pipeline_factory.py
@@ -19,6 +19,9 @@ from pipeline.pipelines import pipeline_base
 # These are required to list the subclasses of pipeline_base
 from pipeline.pipelines import sample_pipeline  # noqa
 from pipeline.pipelines import code_generation_pipeline  # noqa
+from pipeline.tasks import config_tasks
+from pipeline.utils import pipeline_util
+from taskflow.patterns import linear_flow
 
 
 def make_pipeline_flow(pipeline_name, **kwargs):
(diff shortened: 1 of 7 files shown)


jmuk:
cc @tbetbetbe -- I think this would fit into the use case we chatted previously.

jmuk:
config PR https://github.com/googleapis/googleapis/pull/27 is necessary to run the new things properly

pipeline/pipelines/pipeline_factory.py
```diff
@@ -55,3 +58,36 @@ def _rec_subclasses(cls):
...
+        pipeline_name = self.get_pipeline_prefix(**kwargs) + 'ClientPipeline'
+        kwargs['suppress_validate_kwargs'] = True
+        pipeline = make_pipeline(pipeline_name, **kwargs)
+        flow.add(pipeline.flow)
+        return flow
+
+    def get_pipeline_prefix(self, **kwargs):
+        language_prefix = kwargs['language']
+        if language_prefix == 'csharp':
+            language_prefix = 'CSharp'
```
garrettjonesgoogle: This jumped out at me - why do we need to do this capitalization?

jmuk: This is for pipeline name -- most of code generation pipeline is like PythonGapicClientPipeline or JavaGrpcClientPipeline -- <capitalized language name><capitalized client type>ClientPipeline (and an exception is C#, where both C and S are capitalized as CSharpGapicClientPipeline).

PR #23, opened 2016-05-17
Run commands to build package files

jmuk:
This patch introduces tasks for generating package files.
Running setup.py bdist_wheel to generate whl file for Python, and
running rake build to generate gem file for Ruby.

I've also noticed that the 'baseline' directory generated through
merge-task will have a bad side effect on Python binary, so
set up a task to simply copy Gapic codegen results into the
final_repo_dir.


jmuk:
cc: @tbetbetbe @bjwatson @geigerj

tbetbetbe:
LGTM

jmuk:
Okay, I think now it's good to be merged.

PR #33, opened 2016-06-13
Stop building whl file for grpc-python

jmuk:
If output directory is not clean (contains existing files),
packman generates a wrong setup.py and generating whl fails.

The misbehavior of packman is actually due to the side products
of this task, so simply removing task would solve the problem.

Ideally packman should be fixed to ignore unrelated files for
setup.py.


geigerj:
LGTM

PR #34, opened 2016-06-13
Java, PHP: Moving code path prefix from artman to toolkit

garrettjonesgoogle:
Coordinates with https://github.com/googleapis/toolkit/pull/227 .


geigerj:
LGTM

PR #35, opened 2016-06-15
Adding phpcbf into PHP pipeline

garrettjonesgoogle:
This is necessary because phpcbf does different checks from php-cs-fixer, and gcloud-php enforces the rules that phpcbf applies.


shinfan:
LGTM

PR #38, opened 2016-06-17
Convert JavaGrpcClientPipeline to use Packman

mvashishtha:


geigerj:
Nice!

Is the real output of running the the pipeline the same as the old pipeline?

test/testdata/java_grpc_client_pipeline.baseline
```diff
@@ -1,4 +1 @@
-mkdir -p {OUTPUT}/library-v1-gen-java/src/main/java
-protoc --proto_path=test/fake-repos/gapi-core-proto/src/main/proto/ --proto_path=test/fake-repos/fake-proto --proto_path=MOCK_GRADLE_TASK_OUTPUT --java_out={OUTPUT}/library-v1-gen-java/src/main/java test/fake-repos/fake-proto/fake.proto
```
geigerj: This is not strictly related to your PR, but just noting that the old structure provides a bit more flexibility than the new structure. It would be nice to be able to set the `--services_root` and `--include_path` flags for `gen-api-package` so that this task will work on APIs that are not in the googleapis repo.

I've filed #39 for this; it can be handled in a different PR.

mvashishtha: I didn't think about whether the output would be the same at all-- when I test manually, this new version actually produces very different output.

geigerj:
Hm... I'd expect the packman command to generate additional metadata files, but at least the compiled proto files should be the same. Let's confirm that the differences are expected before merging this.

mvashishtha:
@geigerj @ethanbao PTAL. Shin confirmed that running uploadArchives with the build.gradle generated by packman with the "experimental_alt_java" flag produces a maven package that works with gcloud-java. You will need to pull the latest version of packman for this to work.

geigerj:
LGTM

pipeline/pipelines/code_generation_pipeline.py
```diff
@@ -236,8 +236,9 @@ def __init__(self, **kwargs):
 
     def do_build_flow(self, **kwargs):
         flow = linear_flow.Flow('grpc-codegen')
-        flow.add(protoc_tasks.ProtoCodeGenTask('ProtoGen', inject=kwargs),
-                 protoc_tasks.GrpcCodeGenTask('GrpcCodegen', inject=kwargs))
+        kwargs.update({'packman_flags': ['--experimental_alt_java']})
+        flow.add(protoc_tasks.GrpcPackmanTask('Packman', inject=kwargs),
+                 package_tasks.GrpcPackageDirTask('PackageDir', inject=kwargs))
```
ethanbao: I think the original intention for GrpcPackageDirTask is to prepare the package_dir so that one can run package commands. For example, running `setup.py bdist_wheel` inside package_dir to generate whl for Python. @geigerj Can you confirm, Jacob?

Since we don't generate any extra package file for Java, then we don't need this GrpcPackageDirTask.

shinfan:
I don't see any issues in dev mode but it would be nice if we don't change the path. I think /src/main is a convention to put source code in java.

shinfan:
Actually switching from src/main/ to other directory may cause some compile issues in maven.

pipeline/pipelines/code_generation_pipeline.py
geigerj: Good catch. I think you're right; the package dir task appears only to be needed if there is a subsequent task in the pipeline that requires the argument `package_dir`.

ethanbao:
@shinfan maven building is actually fine because we point the source set to /build/generated directory (https://github.com/googleapis/packman/blob/master/templates/java/build-alt.gradle.mustache)

shinfan:
LGTM

mvashishtha:
@ethanbao PTAL. removed package_dir task.

ethanbao:
LGTM

Thanks!

PR #41, opened 2016-06-29
publish maven package for java grpc pipeline if publish_env pipeline keyword is set to 'prod'

mvashishtha:
Currently, because of how the build.gradle template in packman is
configured, only the 'prod' environment (for non-snapshot releases)
is allowed.


pipeline/pipelines/code_generation_pipeline.py
```diff
@@ -35,6 +35,11 @@
                         'pypi_pwd',
                         'publish_env']
 _PYTHON_PUB_ENVS = ['dev', 'test', 'staging', 'prod']
+_JAVA_PUB_REQUIRED = ['maven_repo_url',
+                      'maven_uname',
+                      'maven_pwd',
```
jmuk: I really don't like having there prefixed parameters separately.
This should be consolidated with pypi params, and named simply as 'server_url', 'user_name', and 'password'.

test/testdata/java_grpc_client_pub_pipeline.baseline
```diff
@@ -0,0 +1,2 @@
+gen-api-package --api_name=library/v1 -l java -o {OUTPUT} --experimental_alt_java
+{OUTPUT}/final/gradlew uploadArchives -PmavenRepoUrl=http://example-url/repositories/releases -PmavenUsername=example-maven-uname -PmavenPassword=example-maven-pwd -p{OUTPUT}/final
```
ethanbao: Would that be "-p {OUTPUT}/final" with space in it?

mvashishtha: The flag seems to work without the space as well. Should the space be included?

ethanbao: Never mind, if that works.

test/test_pipeline_baseline.py
```diff
@@ -184,13 +184,29 @@ def test_go_gapic_client_baseline(tmpdir):
                       tmpdir)
 
 
-def _test_java_baseline(task_name, test_name, tmpdir):
-    _test_baseline(task_name, test_name, 'java', str(tmpdir), {})
+def _test_java_baseline(task_name, test_name, tmpdir, extra_kwargs={}):
```
geigerj: [Don't use mutable defaults](http://docs.python-guide.org/en/latest/writing/gotchas/), which can lead to surprising behavior later. Instead do

```
def ...(..., extra_kwargs=None):
  extra_kwargs = extra_kwargs or {}
```

geigerj: Guess I missed this in the Python review as well -- please fix that test, too. (Separately from this PR is OK.)

mvashishtha: Fixed both in this PR.

pipeline/pipelines/code_generation_pipeline.py
mvashishtha: Consolidated python and java parameters.

pipeline/pipelines/code_generation_pipeline.py
```diff
@@ -30,11 +30,11 @@
                   'auto_resolve',
                   'ignore_base',
                   'final_repo_dir']
-_PYTHON_PUB_REQUIRED = ['pypi_server_url',
-                        'pypi_uname',
-                        'pypi_pwd',
-                        'publish_env']
+_PUB_REQUIRED = ['server_url',
+                 'user_name',
+                 'password']
```
ethanbao: Nit: let's name it to "repo_url", "username", "password".

pipeline/pipelines/code_generation_pipeline.py
```diff
@@ -30,11 +30,11 @@
...
 _PYTHON_PUB_ENVS = ['dev', 'test', 'staging', 'prod']
+_JAVA_PUB_ENVS = ['prod']
```
ethanbao: Let's list both staging and prod here.

test/test_pipeline_baseline.py
```diff
@@ -184,13 +185,30 @@ def test_go_gapic_client_baseline(tmpdir):
...
+        'password': 'example-maven-pwd',
+        'publish_env': 'prod'}
+    _test_java_baseline('JavaGrpcClientPipeline',
+                        'java_grpc_client_pub_pipeline',
+                        tmpdir, extra_kwargs)
+
+
+def test_java_grpc_client_nopub_baseline(tmpdir):
+    extra_kwargs = {
+        'server_url': 'http://example-url/repositories/releases',
```
ethanbao: nit: http://maven.example.com/nexus/content/repositories/releases

pipeline/pipelines/code_generation_pipeline.py
mvashishtha: Since the pipeline currently doesn't support snapshot release, should include staging?

pipeline/pipelines/code_generation_pipeline.py
```diff
@@ -105,8 +105,8 @@ def do_build_flow(self, **kwargs):
     def validate_kwargs(self, **kwargs):
         req_args = _VGEN_REQUIRED[:]
         if 'publish_env' in kwargs:
-            req_args += _PYTHON_PUB_REQUIRED
-            _validate_publish_kwargs(**kwargs)
+            req_args += _PUB_REQUIRED
+            _validate_publish_env(_PYTHON_PUB_ENVS, kwargs['publish_env'])
```
jmuk: Can we remove the repeating similar code again here?
Ideally it would be:
```
def validate_kwargs(self, **kwargs):
  req_args = _VGEN_REQUIRED[:]
  _validate_publish_env(req_args, _PYTHON_PUB_ENVS, **kwargs)
  _validate_codegen_kwargs(req_args, **kwargs)
```

Please consider moving more vode into _validate_publish_env to achieve this. This would simplify the code.

pipeline/pipelines/code_generation_pipeline.py
ethanbao: It's different concept. One can have snapshot releases in any environment.

mvashishtha: Okay, I added 'staging'.

test/test_pipeline_baseline.py
mvashishtha: Changed URL

pipeline/pipelines/code_generation_pipeline.py
mvashishtha: I consolidate the publishing validation code.

mvashishtha:
@jmuk @ethanbao PTAL. Added the 'staging' environment and consolidated publishing validation code.

ethanbao:
LGTM

PR #43, opened 2016-06-30
Simplify pipeline execution with presets

shinfan:
For the default tasks you can use a short command:
**python execute_pipeline.py --preset api-language-task** 
Note: now the task can be one of gapic | grpc | core

e.g.
python execute_pipeline.py --preset pubsub-java-gapic

Note: Either pipeline name or preset needs to be provided in the argument to execute the pipeline.

Please let me know what you think. Thanks!

### DIFF
--- a/USAGE.rst
+++ b/USAGE.rst
@@ -111,7 +111,7 @@ Java (pubsub)
      python execute_pipeline.py \
         --config "../googleapis/gapic/core/artman_core.yaml,\
         ../googleapis/gapic/lang/common.yaml" \
-        --language java
+        --language java \
         GrpcClientPipeline
 
      python execute_pipeline.py \


execute_pipeline.py
```diff
@@ -106,22 +114,53 @@ def _parse_args():
...
-  return (flags.pipeline_name,
+  return (pipeline_name,
           pipeline_args,
           flags.remote_mode)
 
 
+def _parse_preset(preset):
+    (api_name, language, task) = preset.strip().split('-')
+    configs = ["../googleapis/gapic/api/artman_%s.yaml:common|%s" % (api_name, language),
+        "../googleapis/gapic/lang/common.yaml:default|%s" % language]
```
garrettjonesgoogle: Knowledge of googleapis is laid out and which pipelines exist should not be in the execute_pipelines script. It should be accessed in an abstract way by execute_pipelines, and encoded next to the actual pipelines.

PR #45, opened 2016-07-12
Make packman always use local protos for gRPC gen

geigerj:
Previously, it always downloaded the public googleapis repo. That does
not work if an API is not on googleapis.


pipeline/tasks/protoc_tasks.py
```diff
@@ -325,6 +325,19 @@ def modify_imports(self, go_import_base, line):
...
 
 class GrpcPackmanTask(packman_tasks.PackmanTaskBase):
-    def execute(self, language, api_name, output_dir, packman_flags=[]):
-        arg_list = [language, api_name, '-o', output_dir] + packman_flags
-        self.run_packman(*arg_list)
+   def execute(self, language, api_name, output_dir, src_proto_path,
+               import_proto_path, packman_flags=[]):
+       arg_list = [language, api_name, '-o', output_dir]
+
+       # TODO: Thos is a workaround for a packman issue:
```
garrettjonesgoogle: spelling(Thos)

geigerj: Fixed

tox.ini
```diff
@@ -21,7 +21,7 @@ commands = flake8 --max-complexity=8 pipeline --exclude='test/output' test
 [testenv:pylint-errors]
 deps = pylint
        -r{toxinidir}/requirements.txt
-commands = pylint -f colorized -E pipeline --ignore test/output test
+commands = pylint -f colorized -E pipeline
```
geigerj: Note: per @michaelbausor we probably were never actually linting this directory, because when you add an `__init__.py` to fix [the build error here](https://travis-ci.org/googleapis/artman/builds/144226454), it produces even more lint errors.

pipeline/tasks/protoc_tasks.py
```diff
@@ -325,6 +325,17 @@ def modify_imports(self, go_import_base, line):
...
+                import_proto_path, packman_flags=[]):
+        api_name = api_name.replace('-', '/')
+        arg_list = [language, api_name, '-o', output_dir]
+
+        def remove_google_dirs(str):
+            return str.split(api_name)[0]
+        imps = map(remove_google_dirs, import_proto_path)
+        srcs = map(remove_google_dirs, src_proto_path)
+
+        arg_list += [arg for src in srcs for arg in ('-r', src)]
```
jmuk: Note that `-r` flag of packman is not repeatable, only the last one has the effect.  Probably better to simply use srcs[0] if it's not empty.

geigerj: Do we have an example where there are multiple paths in `src_proto_path`? I wonder if maybe it would be better to loop over them and execute -- it is unexpected just to ignore the input...

geigerj:
FYI -- I will add PRs to packman and googleapis that correspond to this one soon.

pipeline/tasks/protoc_tasks.py
geigerj: Switch to use a loop.

geigerj:
See https://github.com/googleapis/googleapis/pull/55, googleapis/packman#55

pipeline/tasks/packman_tasks.py
```diff
@@ -23,7 +23,6 @@
 class PackmanTaskBase(task_base.TaskBase):
     def run_packman(self, language, api_name, *additional_args):
         # Fix the api_name convention (ex. logging-v2) for packman.
-        api_name = api_name.replace('-', '/')
```
jmuk: I like to have the same logic into this base class if all of the subclasses are doing same thing.
Why not invoking packman_api_name() here?

pipeline/tasks/protoc_tasks.py
```diff
@@ -325,6 +325,18 @@ def modify_imports(self, go_import_base, line):
...
+
+        def remove_api_dirs(str):
+            return str.split(api_name)[0]
+        imps = map(remove_api_dirs, import_proto_path)
+        srcs = map(remove_api_dirs, src_proto_path)
+
+        arg_list += [arg for imp in imps for arg in ('-i', imp)]
+        for src in srcs:
+            args = arg_list[:]
+            args += ['-r', src] + packman_flags
```
jmuk: Packman's `-r` is not a repeatable flag -- it's expected to be specified once.  I am afraid that it won't work well if srcs have multiple directories.

pipeline/tasks/packman_tasks.py
geigerj: I moved it because it is needed to determine the behavior of `remove_api_dirs` in the gRPC-specific task.

pipeline/tasks/protoc_tasks.py
geigerj: There is only one directory specified here -- it calls packman once for each directory in `srcs`.

jmuk: Ah, you're right, I've missed the point. Thanks.

geigerj:
@jmuk PTAL

pipeline/tasks/protoc_tasks.py
jmuk: Hm, wait a min. We need to make sure that multiple-running packman has no problems.
One thing I noticed is the NodeJS case. It generates 'service.js' which loads specified proto files. Next invocation of packman will overwrite the previous one. There might be something else.

geigerj: Note: per offline discussion, let's alter https://github.com/googleapis/googleapis/pull/55 so that packman can accept multiple `-r` flags. Then we don't have to worry about the overwriting.

jmuk:
LGTM.  Sorry I missed this.

PR #47, opened 2016-07-14
Support artman remote execution. Instead of running artman locally, w…

ethanbao:
…hich requires users to install all dependencies on the dev machine, artman remote execution only requires users to install artman from pypi. This feature is implemented by:

1) Allow user to pass local proto/yaml files to remote artman conductors. The files content (base64 encoded) will be passed via pipeline argument. The mechanism might change in the future, but will be transparent to users.
2) The remote artman conductors (running on docker containers) is going to prepare a temporary directory. It will first download the latest googleapis repo, and then add/update the files in the repo with the ones uploaded from local machine. The temporary directory will be deleted after the execution.
3) The generated output (including the local googleapis repo) will be uploaded to Google Cloud Storage.
4) execute_pipeline.py will post the artman job and wait till the job is done. If the caller has access to the GCS bucket that is used to store uploaded generated output, it will be downloaded to local machine.

Things need to be improved in the followup PRs:
1) Improve the error message when the remote exectution fails.
2) Make sure that nothing gets left on the remote docker containers when the remote execution fails.
3) Add unit and e2e tests for remote execution.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,18 +20,24 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     git
 
 # Install runtime packages.
-RUN apt-get install -y \
+RUN apt-get update && \
+    apt-get install -y \
     ruby \
     python \
     python-dev \
     python-pip \
-    openjdk-7-jre-headless
+    unzip \
+    perl \
+    openjdk-7-jdk
+
+# Define commonly used JAVA_HOME variable
+ENV JAVA_HOME /usr/lib/jvm/java-7-openjdk-amd64
 
 ADD ./scripts /scripts
 WORKDIR /scripts
 # This is need to install nodejs 4.x otherwise nodejs 0.x will be installed.
 RUN bash setup_node4.sh
-RUN apt-get install -y nodejs 
+RUN apt-get install -y nodejs
 
 # Install linuxbrew.
 RUN useradd -m -s /bin/bash linuxbrew
@@ -40,11 +46,12 @@ USER linuxbrew
 WORKDIR /home/linuxbrew
 ENV PATH /home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin:$PATH
 ENV SHELL /bin/bash
-RUN yes |ruby -e "$(curl -fsSL https://raw.github.com/Homebrew/linuxbrew/go/install)"
+RUN yes |ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Linuxbrew/install/master/install)"
 RUN brew doctor || true
 
 # Install protoc and grpc.
-RUN curl -s https://raw.githubusercontent.com/grpc/homebrew-grpc/master/scripts/install | bash -s
+RUN brew tap grpc/grpc
+RUN brew install --with-plugins grpc
 
 # Install Go1.6 from linuxbrew.
 RUN brew install go
@@ -56,7 +63,7 @@ ENV PATH $GOPATH/bin:/home/linuxbrew/.linuxbrew/opt/go/libexec/bin:$PATH
 RUN mkdir -p "$GOPATH/src" "$GOPATH/bin" && chmod -R 777 "$GOPATH"
 
 # Install pacman
-RUN npm install -g googleapis-packman
+RUN npm install -g googleapis-packman@0.7.3
 
 # Setup tools for codegen of Ruby
 RUN gem install rubocop --version '= 0.39.0' --no-ri --no-rdoc
@@ -67,9 +74,10 @@ RUN gem install rake --version '= 10.5.0' --no-ri --no-rdoc
 WORKDIR /
 RUN git clone https://github.com/googleapis/googleapis
 RUN git clone https://github.com/googleapis/toolkit
+ENV TOOLKIT_HOME /toolkit
 
 # Run the pipeline.
-# TODO(cbao): pipeline should be installed from package manager.
+# TODO(ethanbao): pipeline should be installed from package manager.
 ADD . /src
 WORKDIR /src
 RUN pip install -r requirements.txt
--- a/execute_pipeline.py
+++ b/execute_pipeline.py
@@ -37,51 +37,87 @@ Example:
 
 import argparse
 import ast
+import base64
 import os
+import tempfile
+import time
+import uuid
 import yaml
 
-from taskflow import engines, task
+from gcloud import storage
+from gcloud import logging
+from taskflow import engines, task, states
 from pipeline.pipelines import pipeline_factory
-from pipeline.utils import job_util
+from pipeline.utils import job_util, pipeline_util
 
 
 def main():
-  pipeline_name, pipeline_kwargs, remote_mode = _parse_args()
-  pipeline = pipeline_factory.make_pipeline(pipeline_name, **pipeline_kwargs)
+  pipeline_name, pipeline_kwargs, env, local_repo = _parse_args()
 
-  if remote_mode:
-    job_util.post_remote_pipeline_job(pipeline)
+  if local_repo:
+    pipeline_kwargs = _load_local_repo(local_repo, **pipeline_kwargs)
+
+  if env:
+    # Execute pipeline task remotely based on the specified env param.
+    pipeline = pipeline_factory.make_pipeline(
+        pipeline_name, True, **pipeline_kwargs)
+    jb = job_util.post_remote_pipeline_job_and_wait(pipeline, env)
+    task_details, flow_detail = job_util.fetch_job_status(jb, env)
+
+    for task_detail in task_details:
+        if task_detail.name == 'BlobUploadTask' and task_detail.results:
+            bucket_name, path, _ = task_detail.results
+            pipeline_util.download_from_gcs(bucket_name, path,
+                                            os.path.join(tempfile.gettempdir(),
+                                            'artman-remote'))
+
+    if flow_detail.state != 'SUCCESS':
+      # Print the remote log if the pipeline execution completes but not with
+      # SUCCESS status.
+      _print_log(pipeline_kwargs['pipeline_id'])
+
   else:
-    # Hardcoded to execute the pipeline in serial engine, though not necessarily.
-    engine = engines.load(pipeline.flow, engine="serial", store=pipeline.kwargs)
+    pipeline = pipeline_factory.make_pipeline(
+        pipeline_name, False, **pipeline_kwargs)
+    # Hardcoded to run pipeline in serial engine, though not necessarily.
+    engine = engines.load(pipeline.flow, engine='serial', store=pipeline.kwargs)
     engine.run()
 
 
 def _CreateArgumentParser():
   parser = argparse.ArgumentParser()
-  parser.add_argument("pipeline_name",
-                      type=str,
-                      help="The name of the pipeline to run")
   parser.add_argument(
-      "--remote_mode",
-      action="store_true",
-      help="When specified, the pipeline will be executed remotely")
-  parser.add_argument(
-      "--pipeline_kwargs",
+      'pipeline_name',
       type=str,
-      default="{}",
+      help='The name of the pipeline to run')
+  parser.add_argument(
+      '--pipeline_kwargs',
+      type=str,
+      default='{}',
       help=
-      "pipeline_kwargs string, e.g. {\'sleep_secs\':3, \'type\':\'sample\'}")
+      'pipeline_kwargs string, e.g. {\'sleep_secs\':3, \'type\':\'sample\'}')
   parser.add_argument(
-      "--config",
+      '--config',
       type=str,
-      help="Comma-delimited list of the form "
-          + "/path/to/config.yaml:config_section")
+      help='Comma-delimited list of the form '
+          + '/path/to/config.yaml:config_section')
   parser.add_argument(
-      "--reporoot",
+      '--reporoot',
       type=str,
-      default="..",
-      help="Root directory where the input, output, and tool repositories live")
+      default='..',
+      help='Root directory where the input, output, and tool repositories live')
+  parser.add_argument(
+      '--local_repo',
+      type=str,
+      default=None,
+      help='Directory where local proto and gapic configs lives.')
+  parser.add_argument(
+        '--env',
+        type=str,
+        default=None,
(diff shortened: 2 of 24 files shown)


Dockerfile
```diff
@@ -27,11 +27,20 @@ RUN apt-get install -y \
     python-pip \
     openjdk-7-jre-headless
 
+# Install Java.
+RUN \
+    apt-get update && \
+    apt-get install -y openjdk-7-jdk && \
```
jmuk: can we merge this with the previous RUN? this looks like a part of 'Install runtime packages.

execute_pipeline.py
```diff
@@ -137,5 +168,17 @@ def _var_replace(in_str, repl_vars):
   return new_str
 
 
+def _load_local_repo(private_repo_root, **pipeline_kwargs):
+  files_dict = {}
+  for root, dirs, files in os.walk(private_repo_root):
+      for fname in files:
+        path = root+'/'+ fname
```
geigerj: nit: spaces around `+`, or better, use `os.path.join`

pipeline/pipelines/code_generation_pipeline.py
```diff
@@ -97,6 +98,7 @@ def __init__(self, **kwargs):
 
     def do_build_flow(self, **kwargs):
         flow = linear_flow.Flow('grpc-codegen')
+        flow.add(io_tasks.PrepareDirsTask('PrepareDirsTask', inject=kwargs))
```
jmuk: style nit: probably better to keep the consistent style -- use `flow.add(io_task.PrepareDirsTask(...), protoc_tasks....)` rather than invoking flow.add twice.

geigerj:
BTW, see https://github.com/googleapis/artman/pull/45/commits/3a1635aea8a4e8c262ce0dccbf331c0461bb74be regarding the CI build failure.

pipeline/tasks/io_tasks.py
```diff
@@ -0,0 +1,128 @@
...
+
+class CleanupTempDirsTask(task_base.TaskBase):
+    """Clean up all temporary directories"""
+
+    def execute(self, repo_root):
+        if os.path.exists(repo_root) and os.path.isdir(repo_root):
+            shutil.rmtree(repo_root)
+
+
+class PrepareDirsTask(task_base.TaskBase):
```
geigerj: docstring?

pipeline/tasks/io_tasks.py
```diff
@@ -0,0 +1,128 @@
...
+import subprocess
+import shutil
+import urllib
+import zipfile
+
+from gcloud import storage
+from oauth2client.client import GoogleCredentials
+from pipeline.tasks import task_base
+
+DEFAULT_PROJECT = 'vkit-pipeline'
```
jmuk: FYI but 'vkit' is kinda our internal name, might be good to avoid for a default value in the opensource code base.  'gapic-pipeline' might be a good name, although I'm not sure we can take that project name.

pipeline/tasks/io_tasks.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+class PrepareDirsTask(task_base.TaskBase):
+
+    default_provides = ('repo_dir')
+
+    def execute(self, repo_root, files_dict={}):
+        if os.path.exists(repo_root):
+            # Do nothing if the repo_root exists. The repo_root exists if
+            # artman is running locally.
```
michaelbausor: Is it valid for someone running artman locally to specify a repo_root that doesn't exist? This comment indicates it isn't, but in that situation it seems the repo_root will be created and a clean googleapis checked out.

pipeline/tasks/io_tasks.py
```diff
@@ -0,0 +1,128 @@
...
+class BlobDownloadTask(task_base.TaskBase):
+    """A task which downloads file to Google Cloud Storage.
+
+  It requires authentication be properly configured."""
+
+    def execute(self, bucket_name, path, output_dir, project=DEFAULT_PROJECT):
+        client = storage.Client(project=project, credentials=get_credential())
+        bucket = client.get_bucket(bucket_name)
+        blob = bucket.get_blob(path)
+        if blob:
```
jmuk: I personally like the style of:
```
if not blob:
    print 'Cannot find the output from GCS.'
    return
filename = ...
...
```
This reduces the indent.

Also -- is it good to accept nonexisting blobs?  Should we raise an error?

pipeline/tasks/io_tasks.py
jmuk: +1.  Also, "PrepareDirsTask" sounds like too generic to me. Could it be `PrepareGoogleapisTask` or something?

ethanbao: Changed it to PrepareGoogleapisDirTask

pipeline/tasks/io_tasks.py
ethanbao: Good catch. It works locally (and that was intended). I didn't document it to give myself the flexibility to remove this feature from local mode in the future.

If you specify a non-existing directory in the local mode. It will prepare the local googleapis dir and optionally copy the extra proto/yamls files into the temp dir. I found this useful for local mode.

ethanbao:
PTAL. 'codecov/patch' is failing because execute() method for couple of new tasks cannot be well tested under current baseline test frramework. An E2E test framework will be introduced in a followup PR.

execute_pipeline.py
```diff
@@ -82,6 +109,18 @@ def _CreateArgumentParser():
...
+      type=str,
+      default=None,
+      help="Directory where local proto and gapic configs lives. This flag is "
+           "used for remote execution only.")
+  parser.add_argument(
+        "--env",
+        type=str,
+        default="LOCAL",
+        help="Specify which environment should be used to execute pipeline. "
+             "Valid values are 'LOCAL' and 'REMOTE'")
```
jmuk: I'm not sure why you switched from a boolean flag (--remote-mode) to a string flag -- but doing this, you should add choices (see https://docs.python.org/2/library/argparse.html#choices)

execute_pipeline.py
```diff
@@ -104,8 +143,19 @@ def _parse_args():
...
   flags = parser.parse_args()
 
+  repo_root = flags.reporoot
   pipeline_args = {}
-  repl_vars = {"REPOROOT": flags.reporoot}
+
+  if  flags.env.lower() != 'local':
+    pipeline_id = str(uuid.uuid4())
+    # For remote execution, repo_root will be a random temp directory.
+    repo_root = os.path.join('/', 'tmp', 'artman', pipeline_id)
```
jmuk: Better to use mkdtemp: https://docs.python.org/2/library/tempfile.html#tempfile.mkdtemp

execute_pipeline.py
```diff
@@ -37,35 +37,62 @@
...
 
 def main():
-  pipeline_name, pipeline_kwargs, remote_mode = _parse_args()
-  pipeline = pipeline_factory.make_pipeline(pipeline_name, **pipeline_kwargs)
+  pipeline_name, pipeline_kwargs, env, local_repo = _parse_args()
+
+  if local_repo:
+    pipeline_kwargs = _load_local_repo(local_repo, **pipeline_kwargs)
+
+  if env != 'local':
```
jmuk: It's better to avoid negated conditionals. Here I think it's better to say `if env is 'remote'`.

Also -- these 'local' and 'remote' should be declared as constants rather than immediate values.

execute_pipeline.py
```diff
@@ -137,5 +188,31 @@ def _var_replace(in_str, repl_vars):
   return new_str
 
 
+def _load_local_repo(private_repo_root, **pipeline_kwargs):
+  files_dict = {}
+  for root, dirs, files in os.walk(private_repo_root):
+      for fname in files:
+        path = os.path.join(root, fname)
+        rel_path = os.path.relpath(path, private_repo_root)
```
jmuk: This is quite nitpicking, but in case the local environment is Windows, `rel_path` may contain '\' for the path separator and that would confuse the remote system pretty much. I guess it's safer to invoke `os.path.split()` and use the list of path components instead of path string.

jmuk: Or build the tree of dicts based on the file tree structure. @mvashishtha is doing his version for his git-push PR.

pipeline/pipelines/code_generation_pipeline.py
```diff
@@ -11,97 +11,86 @@
...
         kwargs['language'] = 'python'
         super(PythonGrpcClientPipeline, self).__init__(**kwargs)
 
     def do_build_flow(self, **kwargs):
         flow = linear_flow.Flow('grpc-codegen')
-        flow.add(protoc_tasks.GrpcPackmanTask('Packman', inject=kwargs),
-                 package_tasks.GrpcPackageDirTask('PackageDir', inject=kwargs))
+        flow.add(
+            io_tasks.PrepareGoogleapisDirTask('PrepareGoogleapisDirTask',
+                                              inject=kwargs),
```
jmuk: Suggestion:

So it seems all of those code generation pipelines have this task. Can we have a superclass to do that?  Like CodeGenerationPipelineBase, which creates a flow and adds PrepareGoogleapisDirTask. Then, each of the subclass can add their own tasks after that.

Not related to your PR, but some superclasses (like GrpcClientPipelineBase or GapicClientPipelineBase) might reduce the code a bit more.

michaelbausor: I have a PR open here that does a refactoring similar to what you are describing:
https://github.com/googleapis/artman/pull/44

As discussed with @ethanbao the plan was to wait until this PR is merged before going ahead with that one. Perhaps once this PR is merged I can update my PR to reflect the changes you are suggesting?

execute_pipeline.py
ethanbao: I am doing this to support multiple environments. Eventually, we will have limited environment choices. It now can be arbitrary value for better local testing purpose. Developer can start a conductor locally (via `python start_conductor.py --jobboard_name=ethan` and test the change in another console by specify '--env=ethan').

execute_pipeline.py
ethanbao: I change the default value of 'env' flag to be None.

execute_pipeline.py
ethanbao: That is a good point. I use _normalize_path method to turn path into one using slash as separator. I am not using filepath parts list because it is non-hashable and cannot be used a key in dict.

Currently, only relative paths of local files get normalized, and only makes the client side code platform-independent. More work needs to be done if we want to support using Windows for remote conductor.

pipeline/pipelines/code_generation_pipeline.py
ethanbao: @michaelbausor , please help on the refactoring once this PR gets merged.

@jumk, I will leave it as-is in this PR.

ethanbao:
PTAL

pipeline/pipelines/code_generation_pipeline.py
jmuk: got it, makes sense.

Dockerfile
```diff
@@ -40,11 +46,14 @@ USER linuxbrew
...
 ENV PATH /home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin:$PATH
 ENV SHELL /bin/bash
-RUN yes |ruby -e "$(curl -fsSL https://raw.github.com/Homebrew/linuxbrew/go/install)"
+RUN yes |ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Linuxbrew/install/master/install)"
 RUN brew doctor || true
+#RUN brew update
 
 # Install protoc and grpc.
-RUN curl -s https://raw.githubusercontent.com/grpc/homebrew-grpc/master/scripts/install | bash -s
+#RUN curl -s https://raw.githubusercontent.com/grpc/homebrew-grpc/master/scripts/install | bash -s
```
jmuk: Remove this line

execute_pipeline.py
jmuk: So this is going to be remote-mode by default?

Dockerfile
ethanbao: Done.

execute_pipeline.py
```diff
@@ -82,6 +106,18 @@ def _CreateArgumentParser():
...
+      type=str,
+      default=None,
+      help="Directory where local proto and gapic configs lives. This flag is "
+           "used for remote execution only.")
+  parser.add_argument(
+        "--env",
+        type=str,
+        default=None,
+        help="Specify which environment should be used to execute pipeline. "
+             "Valid values are 'LOCAL' and 'REMOTE'")
```
jmuk: The spec of this flag is still unclear to me. Please add more comments on help as well.  What I care are:

- What will happen if it's not specified (`None`)?  It clearly doesn't satisfy `if flags.env` in line 146, so it does not sett up `pipeline_id`, but `env != 'local'` is true at line 60 so it tries to run the remote pipeline. Is it the expected behavior?
- The case sensitiveness -- the help string says 'LOCAL' and 'REMOTE', but they are case-insensitive actually because you invoke tolower(). Please say they are case-insensitive, I prefer lowercase `'local'` and `'remote'` for the example values (if this is going to be case-sensitive after you made other environment -- then I think these "local" and "remote" also should be case-sensitive, it's less confusing).

ethanbao: Thanks for catching this. It is intended to be case-insensitive. Changed line 60 and 172 to fix the bug.

execute_pipeline.py
```diff
@@ -104,8 +140,22 @@ def _parse_args():
...
+    # Use a unique randome temp directory for remote execution.
+    # TODO(ethanbao): Let remote artman decide the temp directory.
+    repo_root = '/tmp/artman/%s' % pipeline_id
+    pipeline_args['pipeline_id'] = pipeline_id
+
+  pipeline_args['repo_root'] = repo_root
+  # TODO(ethanbao): Remove TOOLKIT_HOME var after toolkit_path is removed from
+  # gapic yaml.
+  repl_vars = {"REPOROOT": repo_root,
+               "TOOLKIT_HOME": os.path.join(flags.reporoot, 'toolkit')}
```
jmuk: style nit: use single quotes ('') for normal string -- double quotes for docstring (""" ... """), because it seems that style is common among other files.

jmuk:
One style nit, otherwise LGTM

PR #48, opened 2016-07-15
Stop pylint'ing the test directory

geigerj:
This is causing a build failure currently. We should figure out how
to enable it later. Per @michaelbausor, we probably never were
actually linting it; it was being ignored until a pylint update caused
it to become an error.


ethanbao:
LGTM

PR #49, opened 2016-07-15
Remove bdist_wheel task

geigerj:
This is needed to run `setup.py bdist_wheel` for the Python GAPIC
packman task.


jmuk:
LGTM, but you can remove PythonPackageGenTask instead (or at least remove the reference from PythonGapicClientPipeline).

I thought it was useful when I added it, but I'm not sure how many of us are happy with that -- you know, publishing packages are done without this, and we don't use the generated files quite frequently.

jmuk:
LGTM

PR #50, opened 2016-07-20
add github push task, using util function in pipeline/utils. 

mvashishtha:
I also added a fake repository for testing in test/fake-repos.

Once we had extended the Python and Java GRPC client pipelines in artman to optionally publish generated code, we wanted to extend the Go pipeline to do the same. For Go the standard practice for publishing is simply to upload code to a GitHub repository. Since the task of uploading generated code to a GitHub repository might be useful for other languages as well, we want to add a task that will upload a folder to a GitHub repository (independent of the particular language). Code similar to this will also be used for publishing API protos from google3 to GitHub.  

This commit uses the GitHub API's JSON over HTTP methods to create and push the commit, because libraries that wrap the GitHub API did not provide the required methods.

The implementation follows these steps ([suggested on the API page](https://developer.github.com/v3/git/)), manipulating git internals to publish the output folder:
1. get the current commit object
2. retrieve the tree it points to
3. retrieve the content of the blob object that tree has for that particular file path
4. change the content somehow and post a new blob object with that new content, getting a blob SHA back
5. post a new tree object with that file path pointer replaced with your new blob SHA getting a tree SHA back
6. create a new commit object with the current commit SHA as the parent and the new tree SHA, getting a commit SHA back
7. update the reference of your branch to point to the new commit SHA

### DIFF
--- a/pipeline/tasks/publish_tasks.py
+++ b/pipeline/tasks/publish_tasks.py
@@ -16,6 +16,7 @@
 
 import subprocess
 from pipeline.tasks import task_base
+from pipeline.utils import github_utils
 
 
 class PypiUploadTask(task_base.TaskBase):
@@ -44,6 +45,7 @@ class PypiUploadTask(task_base.TaskBase):
 
 class MavenDeployTask(task_base.TaskBase):
     """Publishes to a Maven repository"""
+
     def execute(self, repo_url, username, password, publish_env,
                 final_repo_dir):
         subprocess.check_call(
@@ -53,6 +55,22 @@ class MavenDeployTask(task_base.TaskBase):
              '-PmavenUsername=' + username,
              '-PmavenPassword=' + password,
              '-p' + final_repo_dir])
+
+    def validate(self):
+        return []
+
+
+class GitHubPushTask(task_base.TaskBase):
+    """Uploads local files to GitHub repository in a new commit. Won't
+    delete files (if missing in local repo) from the remote repo. If the
+    remote copy of a file differs from local copy, overwrites with local copy.
+    Does not change files or folders that are in remote repo but not in local
+    repo."""
+
+    def execute(self, owner, branch, username, password, publish_env,
+                dir_to_push, message):
+        github_utils.push_dir_github(dir_to_push, username, password, owner,
+                                     publish_env, branch, message)
 
     def validate(self):
         return []
--- a/pipeline/tasks/publish_tasks.py
+++ b/pipeline/tasks/publish_tasks.py
@@ -61,12 +61,14 @@ class MavenDeployTask(task_base.TaskBase):
 
 
 class GitHubPushTask(task_base.TaskBase):
-    """Uploads local files to GitHub repository in a new commit. Won't
-    delete files (if missing in local repo) from the remote repo. If the
+    """Uploads local files to GitHub repository in a new commit.
+
+    Won't delete files (if missing in local repo) from the remote repo. If the
     remote copy of a file differs from local copy, overwrites with local copy.
     Does not change files or folders that are in remote repo but not in local
-    repo."""
+    repo.
 
+    """
     def execute(self, owner, branch, username, password, publish_env,
                 dir_to_push, message):
         github_utils.push_dir_github(dir_to_push, username, password, owner,
--- a/pipeline/tasks/publish_tasks.py
+++ b/pipeline/tasks/publish_tasks.py
@@ -67,12 +67,11 @@ class GitHubPushTask(task_base.TaskBase):
     remote copy of a file differs from local copy, overwrites with local copy.
     Does not change files or folders that are in remote repo but not in local
     repo.
-
     """
     def execute(self, owner, branch, username, password, publish_env,
                 dir_to_push, message):
-        github_utils.push_dir_github(dir_to_push, username, password, owner,
-                                     publish_env, branch, message)
+        github_utils.push_dir_to_github(dir_to_push, username, password, owner,
+                                        publish_env, branch, message)
 
     def validate(self):
         return []
--- a/pipeline/utils/github_utils.py
+++ b/pipeline/utils/github_utils.py
@@ -1,0 +1,132 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Utility functions related to GitHub interaction. Uses the GitHub API with
+reference https://developer.github.com/v3/"""
+
+import requests
+import json
+import os
+
+_API_URL = 'https://api.github.com'
+# file modes for different git file types
+_TREE_EL_MODES = {'file': u'100644',
+                  'executable': u'100755',
+                  'tree': u'040000',
+                  'symlink': u'120000'}
+_REQUEST_FUNCS = {'GET': requests.get,
+                  'POST': requests.post,
+                  'PATCH': requests.patch}
+
+
+def _is_executable(el):
+    return os.path.isfile(el) and os.access(el, os.X_OK)
+
+
+def _get_immediate_subdirs(a_dir):
+    return filter(lambda x: os.path.isdir(os.path.join(a_dir, x)),
+                  os.listdir(a_dir))
+
+
+def _get_immediate_files(a_dir):
+    return filter(lambda x: os.path.isfile(os.path.join(a_dir, x)),
+                  os.listdir(a_dir))
+
+
+def _exec_request(http_method, url, req_data=None, username=None,
+                  password=None):
+    """Make an HTTP request with given data and (username, password) tuple.
+    Return json response if request succeeds, otherwise raise an exception"""
+    request_func = _REQUEST_FUNCS[http_method]
+    if username and password:
+        attempt = request_func(url, data=req_data, auth=(username, password))
+    else:
+        attempt = request_func(url, data=req_data)
+    if attempt.ok:
+        return json.loads(attempt.content)
+    else:
+        raise Exception(
+            'Failed request at {0}. Reason: {1}'.format(url, attempt.content))
+
+
+def _get_subtree_sha(orig_tree, path):
+    tree_mode = _TREE_EL_MODES['tree']
+    gen = (x for x in orig_tree['tree'] if (x['path'] == path and
+                                            x['mode'] == tree_mode))
+    subtree_item = next(gen, None)
+    if subtree_item:
+        return subtree_item['sha']
+
+
+# TODO: Have this function also upload symlinks to the repository
+# according to GitHub protocol
+def _extend_git_tree(curr_dir, username, password, base_url,
+                     orig_tree_sha=None):
+    if orig_tree_sha:
+        orig_tree = _exec_request(
+            'GET', base_url + 'git/trees/{0}'.format(orig_tree_sha))
+    tree_els = []
+    for subdir in _get_immediate_subdirs(curr_dir):
+        if orig_tree_sha:
+            subtree_sha = _get_subtree_sha(orig_tree, subdir)
+        else:
+            subtree_sha = None
+        subtree = _extend_git_tree(
+            os.path.join(curr_dir, subdir), username, password, base_url,
+            subtree_sha)
+        tree_els.append({'path': subdir, 'mode': _TREE_EL_MODES['tree'],
+                         'type': 'tree', 'sha': subtree['sha']})
+    for exec_or_file in _get_immediate_files(curr_dir):
+        with open(os.path.join(curr_dir, exec_or_file), 'r') as f:
+            path_content = f.read()
+        mode = _TREE_EL_MODES['file']
+        if _is_executable(os.path.join(curr_dir, exec_or_file)):
+            mode = _TREE_EL_MODES['executable']
+        tree_els.append({'path': exec_or_file, 'mode': mode, 'type': 'blob',
+                         'content': path_content})
+    data_dict = {'tree': tree_els}
+    if orig_tree_sha:
+        data_dict['base_tree'] = orig_tree_sha
+    if not tree_els:
+        raise Exception("Cannot create empty folder {0} in Git".format(
+            curr_dir))
+    return _exec_request('POST', base_url + 'git/trees',
+                         json.dumps(data_dict), username, password)
+
+
+def push_dir_github(output_dir, username, password, owner, repo, branch,
+                    message='Automated commit from artman'):
+    """ Push all content in output_dir to the given GitHub repo branch"""
+    base_url = '{0}/repos/{1}/{2}/'.format(_API_URL, owner, repo)
+    # get the sha of the latest commit on this branch
+    branch_item = _exec_request(
+        'GET',
+        base_url + 'branches/{0}'.format(branch))
+    commit_sha = branch_item['commit']['sha']
+    orig_tree_sha = branch_item['commit']['commit']['tree']['sha']
+    root_tree = _extend_git_tree(output_dir, username, password,
+                                 base_url, orig_tree_sha)
+    # make a new commit using the built tree, with the previous commit as its
+    # only parent
+    req_data = json.dumps({'tree': root_tree['sha'],
+                           'parents': [commit_sha], 'message': message})
+    new_commit_item = _exec_request('POST',
+                                    base_url + 'git/commits', req_data,
+                                    username, password)
+    # update the repository's refs so that the branch head is the new commit
+    _exec_request('PATCH',
+                  base_url + 'git/refs/heads/{0}'.format(branch),
+                  json.dumps({'sha': new_commit_item['sha'],
+                              'force': True}),
+                  username, password)
--- a/pipeline/utils/github_utils.py
+++ b/pipeline/utils/github_utils.py
@@ -46,7 +46,6 @@ def _exec_request(http_method, url, req_data=None, username=None,
 
     Raises:
         HTTPError: HTTP Request failed
-
     """
     if username and password:
         attempt = http_method(url, data=req_data, auth=(username, password))
@@ -91,7 +90,6 @@ def _extend_git_tree(curr_dir, username, password, base_url,
 
     Raises:
         HTTPError: An HTTP request used to construct the tree failed
-
     """
     if orig_tree_sha:
         orig_tree = _exec_request(
@@ -123,9 +121,9 @@ def _extend_git_tree(curr_dir, username, password, base_url,
                              json.dumps(data_dict), username, password)
 
 
-def push_dir_github(output_dir, username, password, owner, repo, branch,
-                    message='Automated commit from artman'):
-    """Push all content in output_dir to the given GitHub repo branch.
+def push_dir_to_github(local_dir, username, password, owner, repo, branch,
+                       message='Automated commit from artman'):
+    """Push all content in local_dir to the given GitHub repo branch.
 
     The effect of running the task is not exactly the same as that of a
     forced git push: this won't change or delete files that are missing
@@ -135,9 +133,8 @@ def push_dir_github(output_dir, username, password, owner, repo, branch,
         None
 
     Raises:
-        ValueError: output_dir was empty
+        ValueError: local_dir was empty
         HTTPError: An HTTP request used to construct the Git tree failed
-
     """
     base_url = '{0}/repos/{1}/{2}/'.format(_API_URL, owner, repo)
     # get the sha of the latest commit on this branch
@@ -146,10 +143,10 @@ def push_dir_github(output_dir, username, password, owner, repo, branch,
         base_url + 'branches/{0}'.format(branch))
     commit_sha = branch_item['commit']['sha']
     orig_tree_sha = branch_item['commit']['commit']['tree']['sha']
-    root_tree = _extend_git_tree(output_dir, username, password,
+    root_tree = _extend_git_tree(local_dir, username, password,
                                  base_url, orig_tree_sha)
     if not root_tree:
-        raise ValueError("Cannot push empty folder {0}".format(output_dir))
+        raise ValueError('Cannot push empty folder {0}'.format(local_dir))
     # make a new commit using the built tree, with the previous commit as its
     # only parent
     req_data = json.dumps({'tree': root_tree['sha'],
--- a/pipeline/utils/github_utils.py
+++ b/pipeline/utils/github_utils.py
@@ -45,7 +45,7 @@ def _exec_request(http_method, url, req_data=None, username=None,
         json response to request
 
     Raises:
-        HTTPError: HTTP Request fails
+        HTTPError: HTTP Request failed
 
     """
     if username and password:
@@ -86,10 +86,11 @@ def _extend_git_tree(curr_dir, username, password, base_url,
     same path, the new tree will use the content of curr_dir.
 
     Returns:
-        New tree object as a dictionary, if curr_dir has files
+        New tree object as a dictionary, if curr_dir had files
         Otherwise, returns None
+
     Raises:
-        HTTPError: An HTTP request used to construct the tree fails
+        HTTPError: An HTTP request used to construct the tree failed
 
     """
     if orig_tree_sha:
@@ -132,9 +133,10 @@ def push_dir_github(output_dir, username, password, owner, repo, branch,
 
     Returns:
         None
+
     Raises:
-        ValueError: curr_dir is empty
-        HTTPError: An HTTP request used to construct the tree fails
+        ValueError: output_dir was empty
+        HTTPError: An HTTP request used to construct the Git tree failed
 
     """
     base_url = '{0}/repos/{1}/{2}/'.format(_API_URL, owner, repo)
--- a/pipeline/utils/github_utils.py
+++ b/pipeline/utils/github_utils.py
@@ -18,52 +18,56 @@ reference https://developer.github.com/v3/"""
 import requests
 import json
 import os
+from urllib2 import HTTPError
 
 _API_URL = 'https://api.github.com'
-# file modes for different git file types
-_TREE_EL_MODES = {'file': u'100644',
-                  'executable': u'100755',
-                  'tree': u'040000',
-                  'symlink': u'120000'}
-_REQUEST_FUNCS = {'GET': requests.get,
-                  'POST': requests.post,
-                  'PATCH': requests.patch}
-
-
-def _is_executable(el):
-    return os.path.isfile(el) and os.access(el, os.X_OK)
+# modes for git objects
+_FILE_MODE = u'100644'
+_EXEC_MODE = u'100755'
+_TREE_MODE = u'040000'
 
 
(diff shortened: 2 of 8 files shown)


geigerj:
The new fake repo is huge... (is it all of artman?). Can we test against a tiny fake repo containing just one or a few files, each file of only one or a few lines?

jmuk:
Why don't we simply invoke git command instead of using github APIs?

ethanbao:
@jmuk Using git command is one option that I once discussed with Mehash before he started implementing. Using git definitely will work for artman but not as simple as maven/pypi is, because the former would require steps like git clone/add/commit/push instead of a single publish step. I would definitely agree that direct use of git command brings lots of benefits. On the other hand, the use of API/library gives us better control and error handling. In addition, the same code will be reused in some other gapic projects where the runtime won't be able to have access to git binary.

mvashishtha:
@geigerj Yeah, it's all of artman. I agree that we don't need something so big for testing. I'll make the repo smaller on my next commit.

mvashishtha:
@geigerj @ethanbao @jmuk  PTAL. Updated the commit message and made the fake repo much smaller.

pipeline/utils/github_utils.py
```diff
@@ -0,0 +1,137 @@
...
+    return {'path': file_or_exec,
+            'mode': _TREE_EL_MODES[blob_mode],
+            'type': 'blob',
+            'sha': blob['sha']}
+
+
+# TODO: Have this function also upload symlinks to the repository
+# according to GitHub protocol
+def _build_git_tree(curr_dir, username, password, owner, repo):
+    """Create a git tree from curr_dir for the given GitHub repo"""
```
jmuk: I think this can be written in a simpler way by using os.walk (see https://docs.python.org/2/library/os.html#os.walk)

pipeline/utils/github_utils.py
```diff
@@ -0,0 +1,137 @@
...
+
+def push_dir_github(output_dir, username, password, owner, repo, branch,
+                    message='Automated commit from artman'):
+    """ Push all content in output_dir to the given GitHub repo branch"""
+    root_tree = _build_git_tree(output_dir, username, password, owner, repo)
+    # get the sha of the latest commit on this branch
+    branch_item = _get_request_content(requests.get,
+                                       '{0}/repos/{1}/{2}/'
+                                       'branches/{3}'.format(_API_URL, owner,
+                                                             repo, branch))
```
jmuk: I like to make a linebreak before 'requests.get', that's better in this case -- such as:
```
branch_item = _get_request_content(
    request.get,
    '{0}/repos/{1}/{2}/branches/{3}'.format(_API_URL, owners, repo, branch))
```

jmuk: Optional: <_API_URL>/repos/<owner>/<repo> is common among this file and other utilities.
I recommend to assign it to a variable, like `base_url = '{0}/repos/{1}/{2}/'.format(_API_URL, owner, repo)` at the beginning of this function, and share it among other invocations.

pipeline/utils/github_utils.py
```diff
@@ -0,0 +1,137 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Utility functions related to GitHub interaction. Uses the GitHub API with
+reference https://developer.github.com/v3/"""
+
+# How to update requirements so that these imports are possible?
```
jmuk: Remove this line

pipeline/utils/github_utils.py
mvashishtha: We discussed this offline-- we decided that because its subtrees have to be created before the tree object is created, os.walk is not appropriate here.

pipeline/tasks/publish_tasks.py
```diff
@@ -56,3 +58,15 @@ def execute(self, repo_url, username, password, publish_env,
...
     def validate(self):
         return []
+
+
+class GitHubPushTask(task_base.TaskBase):
+    """Uploads to a GitHub repository"""
+
+    def execute(self, owner, branch, username, password, publish_env,
+                final_repo_dir, message):
+        github_utils.push_dir_github(final_repo_dir, username, password, owner,
```
ethanbao: Name the first param to "dir_to_push", and use "rebind" (http://docs.openstack.org/developer/taskflow/atoms.html) to map "final_repo_dir" to "dir_to_push" when creating the github publish task. By doing this, we don't force the final_repo_dir to be the input/output name.

pipeline/tasks/publish_tasks.py
```diff
@@ -56,3 +58,15 @@ def execute(self, repo_url, username, password, publish_env,
 
     def validate(self):
         return []
+
+
+class GitHubPushTask(task_base.TaskBase):
+    """Uploads to a GitHub repository"""
```
ethanbao: Let's add details about this task, and emphasize that this task adds and pushes the files in the given directory to a given github repo. It doesn't handle merge and won't delete files (if missing in the local repo) from the remote repo.

pipeline/utils/github_utils.py
```diff
@@ -0,0 +1,137 @@
...
+    return os.path.isfile(el) and os.access(el, os.X_OK)
+
+
+def _get_entries(entry_type_func, a_dir):
+    """List entries in a_dir that are of type specified by entry_type_func"""
+    return filter(lambda x: entry_type_func(os.path.join(a_dir, x)),
+                  os.listdir(a_dir))
+
+
+def _get_request_content(req_func, url, req_data=None, username=None,
```
ethanbao: I feel it more readable if we replace "req_func" param with an enum param (http_method, which can be just GET and POST in this case). For the method name, can we call it "_exec_request", and mention in the comment that this method returns the json response if the request succeeds. Otherwise an exception will be raised.

pipeline/utils/github_utils.py
```diff
@@ -0,0 +1,137 @@
...
+def _build_git_tree(curr_dir, username, password, owner, repo):
+    """Create a git tree from curr_dir for the given GitHub repo"""
+    # Tree_els is a list of children (trees or blobs) for this tree node.
+    tree_els = []
+    for subdir in _get_entries(os.path.isdir, curr_dir):
+        subtree = _build_git_tree(os.path.join(curr_dir, subdir), username,
+                                  password, owner, repo)
+        tree_els.append({'path': subdir,
+                         'mode': _TREE_EL_MODES['tree'],
+                         'type': 'tree', 'sha': subtree['sha']})
```
ethanbao: nit: new line before 'sha'.

pipeline/utils/github_utils.py
```diff
@@ -0,0 +1,137 @@
...
+    for afile in _get_entries(_is_file, curr_dir):
```
ethanbao: nit: let's name "afile" to "file_entry" or "file".

pipeline/utils/github_utils.py
```diff
@@ -0,0 +1,137 @@
...
+
+def _is_file(el):
+    return os.path.isfile(el) and not os.access(el, os.X_OK)
+
+
+def _is_executable(el):
+    return os.path.isfile(el) and os.access(el, os.X_OK)
+
+
+def _get_entries(entry_type_func, a_dir):
```
ethanbao: nit: s/a_dir/dir

pipeline/utils/github_utils.py
```diff
@@ -0,0 +1,137 @@
...
+    else:
+        attempt = req_func(url, data=req_data)
+    if attempt.ok:
+        return json.loads(attempt.content)
+    else:
+        raise Exception('Failed request at {0}.'
+                        'Reason: {1}'.format(url, attempt.content))
+
+
+def _build_git_tree_blob_item(file_or_exec, blob_mode, curr_dir, username,
```
ethanbao: Would that be any difference if the input is file vs executable?

pipeline/utils/github_utils.py
mvashishtha: Both-- done.

pipeline/utils/github_utils.py
mvashishtha: Done.

pipeline/tasks/publish_tasks.py
mvashishtha: Done.

pipeline/tasks/publish_tasks.py
mvashishtha: Done.

pipeline/utils/github_utils.py
mvashishtha: Done.

pipeline/utils/github_utils.py
mvashishtha: Done.

pipeline/utils/github_utils.py
mvashishtha: "afile" no longer used

pipeline/utils/github_utils.py
mvashishtha: dir is a built-in function in python, so I didn't want to override it.

pipeline/utils/github_utils.py
mvashishtha: GitHub requires blobs to have different modes if they are executables rather than files.

mvashishtha:
@ethanbao @jmuk PTAL. 

I added a commit with responses to your initial comments. I made the fake test directories even smaller. I also made the test code simpler. 

I also made a change in functionality after a discussion with Ethan. The previous commit created an entirely new tree, and deleted files that were in the remote repo but not in the local repo. This commit leaves such files alone. It will still overwrite files in the remote repo with content from the local files, however. The commit message has been updated to reflect this.

test/utils/test_github_utils.py
```diff
@@ -0,0 +1,145 @@
...
+
+@mock.patch('pipeline.utils.github_utils._get_immediate_subdirs',
+            side_effect=_imm_subdirs_side_effect)
+@mock.patch('pipeline.utils.github_utils._exec_request',
+            side_effect=_requests_side_effect)
+def test_github_utils_task(mock_exec_request, mock_imm_subdirs):
+    github_utils.push_dir_github(_FAKE_LOCAL, _USERNAME, _PASSWORD, _OWNER,
+                                 _REPO, _BRANCH, _MESSAGE)
+    mock_args = mock_exec_request.call_args_list
+    assert mock_args == EXPECTED_CALL_ARGS
```
jmuk: nit: `mock_exec_request.assert_has_calls(EXPECTED_CALL_ARGS)` would be fine.

jmuk:
LGTM

pipeline/utils/github_utils.py
```diff
@@ -0,0 +1,132 @@
...
+                  'POST': requests.post,
+                  'PATCH': requests.patch}
+
+
+def _is_executable(el):
+    return os.path.isfile(el) and os.access(el, os.X_OK)
+
+
+def _get_immediate_subdirs(a_dir):
+    return filter(lambda x: os.path.isdir(os.path.join(a_dir, x)),
```
geigerj: optional: I find list comprehension more readable here and below:

`[subdir for subdir in os.listdir(a_dir) if os.path.isdir(os.path.join(a_dir, subdir))]`

pipeline/utils/github_utils.py
```diff
@@ -0,0 +1,132 @@
...
+
+def _get_immediate_files(a_dir):
+    return filter(lambda x: os.path.isfile(os.path.join(a_dir, x)),
+                  os.listdir(a_dir))
+
+
+def _exec_request(http_method, url, req_data=None, username=None,
+                  password=None):
+    """Make an HTTP request with given data and (username, password) tuple.
+    Return json response if request succeeds, otherwise raise an exception"""
```
geigerj: Prefer proper google-style doc formatting for return value and exception raising: https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments

pipeline/utils/github_utils.py
```diff
@@ -0,0 +1,132 @@
...
+import json
+import os
+
+_API_URL = 'https://api.github.com'
+# file modes for different git file types
+_TREE_EL_MODES = {'file': u'100644',
+                  'executable': u'100755',
+                  'tree': u'040000',
+                  'symlink': u'120000'}
+_REQUEST_FUNCS = {'GET': requests.get,
```
geigerj: Not sure that this map is really useful -- it requires using string values (e.g., line 104) below where it would be simpler just to pass `requests.XXX` as a parameter directly.

pipeline/utils/github_utils.py
```diff
@@ -0,0 +1,132 @@
...
+def _get_immediate_files(a_dir):
+    return filter(lambda x: os.path.isfile(os.path.join(a_dir, x)),
+                  os.listdir(a_dir))
+
+
+def _exec_request(http_method, url, req_data=None, username=None,
+                  password=None):
+    """Make an HTTP request with given data and (username, password) tuple.
+    Return json response if request succeeds, otherwise raise an exception"""
+    request_func = _REQUEST_FUNCS[http_method]
```
geigerj: See comment above. Would prefer `http_method` were just in `[requests.get, requests.post, requests.patch]` so that you don't need to translate string params

pipeline/utils/github_utils.py
```diff
@@ -0,0 +1,132 @@
...
+    if username and password:
+        attempt = request_func(url, data=req_data, auth=(username, password))
+    else:
+        attempt = request_func(url, data=req_data)
+    if attempt.ok:
+        return json.loads(attempt.content)
+    else:
+        raise Exception(
```
geigerj: Instead of using the Exception base class, raise an appropriate subclass. It's fine to create one if it doesn't exist.

pipeline/utils/github_utils.py
```diff
@@ -0,0 +1,132 @@
...
+"""Utility functions related to GitHub interaction. Uses the GitHub API with
+reference https://developer.github.com/v3/"""
+
+import requests
+import json
+import os
+
+_API_URL = 'https://api.github.com'
+# file modes for different git file types
+_TREE_EL_MODES = {'file': u'100644',
```
geigerj: Instead of dict, why not just `FILE_MODE`, `EXECUTABLE_MODE`, ...?

pipeline/utils/github_utils.py
```diff
@@ -0,0 +1,132 @@
...
+    tree_els = []
+    for subdir in _get_immediate_subdirs(curr_dir):
+        if orig_tree_sha:
+            subtree_sha = _get_subtree_sha(orig_tree, subdir)
+        else:
+            subtree_sha = None
+        subtree = _extend_git_tree(
+            os.path.join(curr_dir, subdir), username, password, base_url,
+            subtree_sha)
+        tree_els.append({'path': subdir, 'mode': _TREE_EL_MODES['tree'],
```
geigerj: See comment above about `_TREE_EL_MODES`

pipeline/utils/github_utils.py
```diff
@@ -0,0 +1,132 @@
...
+    gen = (x for x in orig_tree['tree'] if (x['path'] == path and
+                                            x['mode'] == tree_mode))
+    subtree_item = next(gen, None)
+    if subtree_item:
+        return subtree_item['sha']
+
+
+# TODO: Have this function also upload symlinks to the repository
+# according to GitHub protocol
+def _extend_git_tree(curr_dir, username, password, base_url,
```
geigerj: This function is non-trivial -- could you please add a short docstring describing it, even though it's private?

pipeline/utils/github_utils.py
```diff
@@ -0,0 +1,132 @@
...
+        mode = _TREE_EL_MODES['file']
+        if _is_executable(os.path.join(curr_dir, exec_or_file)):
+            mode = _TREE_EL_MODES['executable']
+        tree_els.append({'path': exec_or_file, 'mode': mode, 'type': 'blob',
+                         'content': path_content})
+    data_dict = {'tree': tree_els}
+    if orig_tree_sha:
+        data_dict['base_tree'] = orig_tree_sha
+    if not tree_els:
+        raise Exception("Cannot create empty folder {0} in Git".format(
```
geigerj: Don't raise base Exception -- pick or create an appropriate subclass.

pipeline/utils/github_utils.py
```diff
@@ -0,0 +1,132 @@
...
+            curr_dir))
+    return _exec_request('POST', base_url + 'git/trees',
+                         json.dumps(data_dict), username, password)
+
+
+def push_dir_github(output_dir, username, password, owner, repo, branch,
+                    message='Automated commit from artman'):
+    """ Push all content in output_dir to the given GitHub repo branch"""
```
geigerj: extraneous space before docstring (`""" P` --> `"""P`)

pipeline/utils/github_utils.py
mvashishtha: Should I list the args in the docstring as well?

pipeline/utils/github_utils.py
mvashishtha: Originally I was passing function objects directly but Ethan thought that enums would be more readable, and I thought that this approach with a dictionary would be more like using enums. I will talk with Ethan about this-- even if we use the Python enum class I think that passing the function objects themselves will be more readable.

pipeline/utils/github_utils.py
mvashishtha: Done for both functions

pipeline/utils/github_utils.py
mvashishtha: Done

pipeline/utils/github_utils.py
mvashishtha: Changed back to passing functions

pipeline/utils/github_utils.py
mvashishtha: Resolved above

pipeline/utils/github_utils.py
mvashishtha: Changed to HTTPError

pipeline/utils/github_utils.py
mvashishtha: Done

pipeline/utils/github_utils.py
mvashishtha: Done

pipeline/utils/github_utils.py
mvashishtha: This error case was moved to push_dir_github and raises a ValueError

pipeline/utils/github_utils.py
mvashishtha: Done

mvashishtha:
@geigerj @ethanbao PTAL. Responded to comments. Also made other changes (mentioned in commit message)

test/utils/test_github_utils.py
```diff
@@ -0,0 +1,147 @@
...
+def _imm_subdirs_side_effect(curr_dir):
+    """Must mock this function so order of subdirectories is known"""
+    if curr_dir == _FAKE_LOCAL:
+        return ['main', 'util']
+    elif curr_dir == os.path.join(_FAKE_LOCAL, 'main'):
+        return ['subdir']
+    return []
+
+
+@mock.patch('pipeline.utils.github_utils._get_immediate_subdirs',
```
geigerj: Don't mock the pipeline -- we want to test this. Instead mock at the lowest level you can -- here, I think that would be the `requests` library.

test/utils/test_github_utils.py
```diff
@@ -0,0 +1,147 @@
...
+                                            'parents': ['fake-commit-sha'],
+                                            'message': 'fake-message'}),
+                                _USERNAME, _PASSWORD),
+                      mock.call(requests.patch, _PATCH_REF_URL,
+                                json.dumps({'force': True,
+                                            'sha': 'fake-commit-sha'}),
+                                _USERNAME, _PASSWORD)]
+
+
+def _requests_side_effect(http_method, url, req_data=None, username=None,
```
geigerj: As discussed offline, I recommend not representing the remote tree, because it requires a substantial amount of complexity in this function -- ideally, the test should have little logic of its own.

Instead, can we mock `requests` and explicitly specify each response to each request, with no reference to the remote tree as a whole?

pipeline/tasks/publish_tasks.py
```diff
@@ -56,3 +58,19 @@ def execute(self, repo_url, username, password, publish_env,
 
     def validate(self):
         return []
+
+
+class GitHubPushTask(task_base.TaskBase):
+    """Uploads local files to GitHub repository in a new commit. Won't
```
geigerj: Nit: prefer to start docstrings with a single summary line, and give details in the following paragraph.

pipeline/utils/github_utils.py
```diff
@@ -0,0 +1,163 @@
...
+
+def _exec_request(http_method, url, req_data=None, username=None,
+                  password=None):
+    """Make an HTTP request with given data and (username, password) tuple.
+
+    Returns:
+        json response to request
+
+    Raises:
+        HTTPError: HTTP Request fails
```
geigerj: Use past tense here and elsewhere ("The HTTP request failed.")

pipeline/utils/github_utils.py
```diff
@@ -0,0 +1,163 @@
...
+    If orig_tree_sha is provided, the new tree inherits all structure from
+    the remote tree with sha orig_tree_sha. The new tree will always include
+    files or paths that are present in the remote tree. However, if the
+    curr_dir and the remote tree  have conflicting contents for files with the
+    same path, the new tree will use the content of curr_dir.
+
+    Returns:
+        New tree object as a dictionary, if curr_dir has files
+        Otherwise, returns None
+    Raises:
```
geigerj: Use a newline between `Returns:` and `Raises:` blocks

pipeline/utils/github_utils.py
```diff
@@ -0,0 +1,163 @@
...
+    """Push all content in output_dir to the given GitHub repo branch.
+
+    The effect of running the task is not exactly the same as that of a
+    forced git push: this won't change or delete files that are missing
+    in the local repo but are present in the remote repo.
+
+    Returns:
+        None
+    Raises:
+        ValueError: curr_dir is empty
```
geigerj: There's no parameter to this function called `curr_dir`

test/utils/test_github_utils.py
mvashishtha: The function that I'm testing has to use os functions to walk through the specified local directory, and I have to mock the function for listing directory items so that I know what order the tree will be built in. What do you mean by mocking the pipeline?

geigerj: I mean that it is better to do

```
@mock.patch('os.path.isdir')
@mock.patch('os.path.listdir')
@mock.patch('requests....')
```

instead of mocking the same code that you're testing -- these utility methods are untested if you're mocking them.

jmuk: I agree that this does not mock the pipeline -- but I am not sure if mocking your own internal function is the best approach.

Probably it is better to mock a filesystem itself (with pyfakefs https://github.com/jmcgeheeiv/pyfakefs) and mock the HTTP functions.

geigerj: Sorry if I was unclear -- by "mock the pipeline", I just meant mocking functions that are defined in the artman repo, as opposed to its dependencies.

jmuk: Yeah, I see (I wrote my reply before I see your 2nd comment).  pyfakefs looks like the utility to mock `os.path` functions consistently.

mvashishtha: Test now uses pyfakefs and mocks at os level.

test/utils/test_github_utils.py
mvashishtha: Now explicitly specifying response to each request.

pipeline/tasks/publish_tasks.py
mvashishtha: Changed docstrings to this pattern.

pipeline/utils/github_utils.py
mvashishtha: Done

pipeline/utils/github_utils.py
mvashishtha: Done

pipeline/utils/github_utils.py
mvashishtha: Changed to output_dir

mvashishtha:
@geigerj @jmuk PTAL. Tests now mock at os and requests level as suggested. I use a fake directory structure with pyfakefs, so I deleted the fake local repo. I also responded to Jacob's comments on the docstrings.

pipeline/tasks/publish_tasks.py
```diff
@@ -56,3 +58,21 @@ def execute(self, repo_url, username, password, publish_env,
...
+
+
+class GitHubPushTask(task_base.TaskBase):
+    """Uploads local files to GitHub repository in a new commit.
+
+    Won't delete files (if missing in local repo) from the remote repo. If the
+    remote copy of a file differs from local copy, overwrites with local copy.
+    Does not change files or folders that are in remote repo but not in local
+    repo.
+
```
geigerj: Nit: no blank line before the closing `"""` (here and elsewhere)

pipeline/utils/github_utils.py
```diff
@@ -0,0 +1,165 @@
...
+    # get the sha of the latest commit on this branch
+    branch_item = _exec_request(
+        requests.get,
+        base_url + 'branches/{0}'.format(branch))
+    commit_sha = branch_item['commit']['sha']
+    orig_tree_sha = branch_item['commit']['commit']['tree']['sha']
+    root_tree = _extend_git_tree(output_dir, username, password,
+                                 base_url, orig_tree_sha)
+    if not root_tree:
+        raise ValueError("Cannot push empty folder {0}".format(output_dir))
```
geigerj: nit: prefer single quotes for strings for consistency with the rest of the codebase

geigerj:
Two nits, then LGTM

tox.ini
```diff
@@ -16,7 +16,7 @@ commands = py.test --timeout=30 --cov-report html --cov-report=term --cov {toxin
 
 [testenv:pep8]
 deps = flake8
-commands = flake8 --max-complexity=8 pipeline --exclude='test/output' test
+commands = flake8 --max-complexity=8 pipeline --exclude='test/output','test/fake-repos' test
```
geigerj: Do you still need this since you are now mocking the file system?

pipeline/utils/github_utils.py
```diff
@@ -0,0 +1,165 @@
...
+    # Git doesn't allow empty trees
+    if tree_els:
+        data_dict = {'tree': tree_els}
+        if orig_tree_sha:
+            data_dict['base_tree'] = orig_tree_sha
+        return _exec_request(requests.post, base_url + 'git/trees',
+                             json.dumps(data_dict), username, password)
+
+
+def push_dir_github(output_dir, username, password, owner, repo, branch,
```
ethanbao: nit: push_dir_to_github, and rename the first param to dir.

ethanbao:
LGTM

pipeline/tasks/publish_tasks.py
mvashishtha: Done

pipeline/utils/github_utils.py
mvashishtha: Changed to single quotes

tox.ini
mvashishtha: Not needed anymore, with fake file system. Removed this part.

pipeline/utils/github_utils.py
mvashishtha: Since dir is reserved name in python, renamed output_dir to local_dir. Renamed push_dir_github to push_dir_to_github.

PR #53, opened 2016-08-08
Skips some unnecessary tasks for Ruby/NodeJS pipeline for gcloud.

jmuk:
- I thought RubyFormatTask was helpful, but it seems not so much,
  and automatic correction might end up with a weird style without
  proper configurations. Thus this task will be removed.
- packman is not necessary for Ruby and NodeJS Gapic pipeline
  if it's a part of gcloud.


pipeline/pipelines/code_generation_pipeline.py
```diff
@@ -207,9 +205,8 @@ def do_build_flow(self, **kwargs):
                                           inject=kwargs),
             gapic_tasks.GapicCodeGenTask('GapicCodegen',
                                          inject=kwargs),
-            gapic_tasks.GapicCopyTask('GapicCopy', inject=kwargs),
-            gapic_tasks.GapicPackmanTask('GapicPackman',
-                                         inject=kwargs))
+            gapic_tasks.GapicCopyTask('GapicMerge', inject=kwargs),
```
geigerj: Why does the parameter name not match the class name here?

pipeline/tasks/gapic_tasks.py
```diff
@@ -165,6 +165,11 @@ def execute(self, gapic_api_yaml):
 
 class GapicPackmanTask(packman_tasks.PackmanTaskBase):
     def execute(self, language, api_name, final_repo_dir):
+        # Some APIs will be a part of gcloud project for Ruby and NodeJS.
+        # Such APIs don't need packman.
+        if ((language is 'ruby' or language is 'nodejs') and
```
geigerj: Is `is` brittle here because it's testing pointer equality rather than content equality?

pipeline/pipelines/code_generation_pipeline.py
jmuk: My bad, this was wrong. Fixed -- using GapicMerge task.

pipeline/tasks/gapic_tasks.py
jmuk: Fixed, you're right.

ethanbao:
LGTM, please fix the check failure.

geigerj:
LGTM once tests pass

PR #59, opened 2016-08-10
Switching JavaCorePipeline to use Packman

garrettjonesgoogle:
Also, using a unique output directory per api

I added support in Packman for this case in commit https://github.com/googleapis/packman/commit/d0209d8d45d24d16547aadfb502dc9b9b78de054 .


garrettjonesgoogle:
Ok tests fixed, PTAL

michaelbausor:
Change looks good. Two things:
- We don't have any tests for the core pipelines. Maybe we should add them?
- This will conflict with https://github.com/googleapis/artman/pull/58, is it urgent to submit this immediately?

garrettjonesgoogle:
* Yeah we should definitely add tests for the core pipelines.
* It's not urgent to submit immediately because I can still run artman locally.

I will wait for your test restructuring, and after I pull that in then I'll add a test for the core java pipeline (at least).

michaelbausor:
OK sounds good

garrettjonesgoogle:
rebased, PTAL

pipeline/pipelines/core_generation_pipeline.py
```diff
@@ -17,20 +17,21 @@
...
 
 from pipeline.pipelines import code_generation_pipeline
+from pipeline.pipelines import grpc_generation_pipeline
 from pipeline.tasks import protoc_tasks
+from pipeline.tasks import publish_tasks
 
 
-class JavaCoreProtoPipeline(
-        code_generation_pipeline.CodeGenerationPipelineBase):
+class JavaCoreProtoPipeline(grpc_generation_pipeline.GrpcClientPipeline):
```
michaelbausor: Add a comment explaining why this inherits from the grpc pipeline. Also maybe consider inheriting directly from JavaGrpcClientPipeline (and updating the packman_flags)

I can see arguments for no inheritance from grpc, inheriting from the base grpc pipeline (as you have done), and inheriting from the java grpc pipeline, so in any case I think some explanation would be good.

garrettjonesgoogle:
Added doc, PTAL

michaelbausor:
LGTM once Travis completes

PR #63, opened 2016-08-18
Java generation: switching from merging to copying

garrettjonesgoogle:


pipeline/tasks/gapic_tasks.py
```diff
@@ -105,15 +105,24 @@ def validate(self):
...
-            'cp', '-rf', gapic_code_dir, final_repo_dir])
+
+
+class GapicCopyTask(task_base.TaskBase):
+    """Copy the files generated by Gapic toolkit to the final_repo_dir.
+    """
+
+    def execute(self, final_repo_dir, gapic_code_dir):
+        print "Copying " + gapic_code_dir + "/* to " + final_repo_dir
+        for entry in os.listdir(gapic_code_dir):
```
michaelbausor: Why is it necessary to do this, rather than the `self.exec_command(['cp', '-rf', gapic_code_dir, final_repo_dir])` in the original method?

garrettjonesgoogle: Because the following command:

`cp -rf ../artman/output/google-cloud-speech-v1-gapic-gen-java ../gcloud-java/gcloud-java-speech`

results in a directory `../gcloud-java/gcloud-java-speech/google-cloud-speech-v1-gapic-gen-java`

But what I want is the contents of `../artman/output/google-cloud-speech-v1-gapic-gen-java` copied over.

michaelbausor: But won't this interfere with python, which presumably wants the directory?

garrettjonesgoogle: I already interfered with python in a previous PR when I renamed this output directory... Anyway, python's final output dir is under output/ , so it currently has to be manually copied.

michaelbausor: OK, fair enough

michaelbausor:
LGTM

PR #64, opened 2016-08-18
Fix bugs in Python GRPC/Gapic package publishing

mvashishtha:
devpi upload failed during manual test by Jacob because the --from-dir
flag for "devpi upload" requires archive files
(see http://doc.devpi.net/latest/userman/devpi_commands.html#upload).
Had subprocess change into ouptut directory before running "devpi
upload".

Also deleted the GrpcPackageDirTask and GapicPackageDirTask
from package_tasks: now package_dir is set by GrpcPackmanTask or
GapicPackmanTask with the default_provides keyword. Maven and PyPI
upload is now done from package_dir rather than final_repo_dir.

Also made a trailing slash for repo_url optional.

### DIFF
--- a/pipeline/tasks/publish_tasks.py
+++ b/pipeline/tasks/publish_tasks.py
@@ -14,6 +14,9 @@
 
 """Tasks for publishing artman output"""
 
+import os
+import urlparse
+
 from pipeline.tasks import task_base
 from pipeline.utils import github_utils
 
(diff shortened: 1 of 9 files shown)


mvashishtha:
@jmuk @geigerj @ethanbao PTAL.

jmuk:
LGTM

pipeline/tasks/publish_tasks.py
```diff
@@ -22,21 +24,21 @@ class PypiUploadTask(task_base.TaskBase):
     """Publishes a PyPI package"""
 
     def execute(self, repo_url, username, password, publish_env,
-                final_repo_dir):
-        publish_url = repo_url + username + '/' + publish_env
+                package_dir):
+        if repo_url[-1] != '/':
```
geigerj: nit: prefer `os.sep`, which is platform agnostic, instead of `'/'`

geigerj:
LGTM with one nit

pipeline/tasks/publish_tasks.py
jmuk: That's not `os.sep` actually, it is always `'/'` because it's for URL.

jmuk: ah, but it's nicer to use `urlparse.urljoin` function.

geigerj: Ah, whoops, you're right.

mvashishtha:
Fixed nit before merge.

PR #67, opened 2016-08-19
Removing merge support, improving test failure messages

garrettjonesgoogle:
Test failures now provide lines that can be copied into baselines,
because they print with {CWD} and {OUTPUT}.


test/test_pipeline_baseline.py
```diff
@@ -43,30 +43,51 @@ def get_expected_calls(baseline, binding, stderr=False):
...
+    'go': 'my_api.go',
+    'ruby': 'my_api.rb',
+    'php': 'MyApi.php',
+    'csharp': 'MyApi.cs',
+    'nodejs': 'my_api.js'
+}
+
+
+def make_fake_gapic_output(output_dir, language):
+    # Create an empty 'README.md' in the output_dir. Do not invoke
```
michaelbausor: Why README.md?

garrettjonesgoogle: oops some documentation may be out of date... it actually uses a filename from the table right above. I'll fix the comment.

test/test_pipeline_baseline.py
```diff
@@ -43,30 +43,51 @@ def get_expected_calls(baseline, binding, stderr=False):
...
+    mismatch_text = 'Mismatch between expected and actual subprocess calls.\n'
+
+    def format_err(exp, act):
+        return mismatch_text + 'Expected:\n' + exp + '\nActual:\n' + act
+    exp = format_calls(expected_calls, bindings)
+    act = format_calls(actual_calls, bindings)
+    assert exp == act, format_err(exp, act)
+
+
+_DUMMY_FILE_DICT = {
```
michaelbausor: Doesn't need to be done in this PR, but what do you think about just adding each of these files to git, and skipping the creation process altogether?

garrettjonesgoogle: I'm fine with that. I vote on that being done in a later PR.

pipeline/pipelines/gapic_generation_pipeline.py
```diff
@@ -98,7 +91,7 @@ def _get_gapic_package_tasks(self, **kwargs):
 class _NodeJSGapicTaskFactory(GapicTaskFactoryBase):
 
     def _get_gapic_package_tasks(self, **kwargs):
-        return [gapic_tasks.GapicMergeTask('GapicMerge', inject=kwargs),
+        return [gapic_tasks.GapicCopyTask('GapicCopy', inject=kwargs),
                 gapic_tasks.GapicPackmanTask('GapicPackman', inject=kwargs)]
```
michaelbausor: Is there a good reason to keep the CSharp task as it is? Or can we make it use the default copy step? (Because the default previously was a merge step, it didn't seem appropriate to change CSharp to use it)

michaelbausor:
LGTM, nits only

pipeline/pipelines/gapic_generation_pipeline.py
garrettjonesgoogle: I think the default copy step would be safe. For 'csharp', final_repo_dir is always set to .../artman/ouput/... . I'll update it.

garrettjonesgoogle:
Nits fixed.

michaelbausor:
LGTM

PR #68, opened 2016-08-19
Create GapicClientBatchPipeline

michaelbausor:
Creates a new pipeline called GapicClientBatchPipeline which generates multiple API and language combinations.

This pipeline uses a new config file at `googleapis/gapic/batch/staging.yaml` (see https://github.com/googleapis/googleapis/pull/96)

Generation by the batch task can be toggled in the artman yaml file for each API, and in the artman language yaml for each language.

This PR also includes:
- Modifying task factories to return a list of task classes, which are instantiated separately
- Simplifies creation of dummy files in test


geigerj:
Let's sync next week and make sure that we're not both working on the same things at once: I had anticipated putting behavior like this into the artman wrapper (although not specifically targeting the staging).

Also, I had expected the more substantial work in batching to be going into the `toolkit` project, rather than artman, so that we're not spinning up a new instance of the GAPIC generator for each language X api.

michaelbausor:
OK sounds good, let's catch up next week.

USAGE.rst
```diff
@@ -235,6 +235,20 @@ Node.JS (logging)
         GapicClientPipeline
 
 
+Gapic Batch Pipeline
+********************
+
+The GAPIC batch pipeline can be used to generate multiple API and language
+combinations with a artman command. This example uses the staging repository
```
garrettjonesgoogle: "a artman command" -> "one artman command"

pipeline/pipelines/batch_generation_pipeline.py
```diff
@@ -0,0 +1,138 @@
...
+
+            single_flow = linear_flow.Flow('SingleLanguageApiFlow')
+            single_flow.add(*code_gen.instantiate_tasks(tasks, api_kwargs))
+            yield single_flow
+
+    def get_pipeline_tasks(self, **kwargs):
+        """Abstract method, subclasses must implement this method and return
+        a list of tasks.
+        """
+        raise NotImplementedError('Subclass must implement abstract method')
```
garrettjonesgoogle: Could this just be provided in the constructor instead?

pipeline/pipelines/gapic_generation_pipeline.py
```diff
@@ -52,20 +51,26 @@ def __init__(self, **kwargs):
             **kwargs)
 
 
+class GapicClientBatchPipeline(batch_gen.BatchPipeline):
+
+    def get_pipeline_tasks(self, **kwargs):
+        task_factory = get_gapic_task_factory(kwargs['language'])
+        return task_factory._get_gapic_codegen_tasks(**kwargs)
```
garrettjonesgoogle: I think we need to do the following:

1. Separate the pipelines for codegen from packaging (not just separate them for the batch case). 
2. Decouple the command line from the implementing classes inside artman; we should come up with identifiers to use, which would determine which pipeline(s) would be created. 

This might be a good surface:

```
# perform codegen for one language/API combo
artman gapic-codegen --language java --api pubsub

# batch APIs, one language
artman gapic-codegen --language java --api all

# batch languages, one API
artman gapic-codegen --language all --api pubsub

# perform packaging
artman gapic-package --language java --api pubsub

# perform codegen + packaging
artman gapic-codegen gapic-package --language java --api pubsub

# perform batch codegen + packaging for all languages and APIs
artman gapic-codegen gapic-package --language all --api all

```

geigerj: Let's talk about this this afternoon. I have already begun working on a new command-line interface for artman, and this work is running up against that.

michaelbausor:
Have updated to address PR feedback, added remote mode support for the GapicClientBatchPipeline, and rebased to fix merge conflicts, PTAL

pipeline/pipelines/batch_generation_pipeline.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def __init__(self, get_pipeline_tasks_func, **kwargs):
+        task_factory = BatchTaskFactory(get_pipeline_tasks_func)
+        super(BatchPipeline, self).__init__(
+            task_factory, **kwargs)
+
+    def validate_kwargs(self, **kwargs):
+        required = ['apis', 'languages', 'api_config_pattern',
+                    'artman_language_yaml', 'batch_output_dir']
+        pipeline_util.validate_exists(required, **kwargs)
```
garrettjonesgoogle: So the parent class (CodeGenerationPipelineBase) has an existing validate_kwargs which calls pipeline_util.validate_exists itself (indirectly), and it gets extra kwargs to validate from the task factory. Couldn't you provide the kwargs to validate through get_validate_kwargs? I know that _validate_codegen_kwargs in code_generation_pipeline has some kwargs that are not relevant to the BatchPipeline, but maybe those should just be moved to the task factory(s) for the non-batch pipelines, and not have CodeGenerationPipelineBase assume anything.

pipeline/pipelines/batch_generation_pipeline.py
```diff
@@ -0,0 +1,133 @@
...
+        flow.add(batch_flow)
+        flow.add(*task_utils.instantiate_tasks(final_tasks, kwargs))
+
+        return flow
+
+    def get_language_api_flows(self, apis, languages, api_config_pattern,
+                               artman_language_yaml, **kwargs):
+        # Remove language arg from kwargs - we don't want it to override
+        # the language set in _get_configs
+        del kwargs['language']
```
garrettjonesgoogle: Should this ever be added back to kwargs?

pipeline/pipelines/batch_generation_pipeline.py
```diff
@@ -0,0 +1,133 @@
...
+    else:
+        return languages.split(',')
+
+
+def _get_api_configs(api_config_pattern, apis):
+    if apis == '*':
+        return sorted(glob.glob(api_config_pattern))
+    else:
+        return [config_util.var_replace(api_config_pattern, {'APINAME': api})
+                for api in apis.split(',')]
```
garrettjonesgoogle: We could validate the api values based on what artman configs are present (this could be a follow-up enhancement)

pipeline/pipelines/batch_generation_pipeline.py
```diff
@@ -0,0 +1,133 @@
...
+
+def _get_languages(artman_language_yaml, languages):
+    if languages == '*':
+        with open(artman_language_yaml) as config_file:
+            languages = sorted(yaml.load(config_file).keys())
+        if 'common' in languages:
+            languages.remove('common')
+        return languages
+    else:
+        return languages.split(',')
```
garrettjonesgoogle: We could validate the language values (this could be a follow-up enhancement)

pipeline/pipelines/batch_generation_pipeline.py
michaelbausor: SGTM, will implement.

pipeline/pipelines/batch_generation_pipeline.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def _get_api_configs(api_config_pattern, apis):
```
garrettjonesgoogle: maybe name this _get_artman_configs instead. We have two types of configs - artman & GAPIC; it's hard to determine what "api config" means in relation to those.

garrettjonesgoogle: Actually, these are file names - maybe "_get_artman_config_filenames"

pipeline/pipelines/batch_generation_pipeline.py
michaelbausor: No, it shouldn't be used anywhere when calling the batch pipeline.

garrettjonesgoogle: Then should it be an error if it's set?

pipeline/pipelines/batch_generation_pipeline.py
```diff
@@ -0,0 +1,133 @@
...
+            single_flow.add(*tasks)
+            yield single_flow
+
+
+def _get_configs(api_config_yamls, languages, artman_language_yaml, repo_root):
+    for language in _get_languages(artman_language_yaml, languages):
+        lang_config = _load_artman_config(artman_language_yaml,
+                                          language,
+                                          repo_root)
+        lang_config['language'] = language
```
garrettjonesgoogle: It feels very weird to do this here. Why not modify api_kwargs directly, further down?

pipeline/pipelines/batch_generation_pipeline.py
```diff
@@ -0,0 +1,133 @@
...
+                 batch_tasks.BatchCleanTask,
+                 batch_tasks.BatchCopyTask],
+                api_kwargs)
+
+            single_flow = linear_flow.Flow('SingleLanguageApiFlow')
+            single_flow.add(*tasks)
+            yield single_flow
+
+
+def _get_configs(api_config_yamls, languages, artman_language_yaml, repo_root):
```
garrettjonesgoogle: Maybe name this "_get_api_kwarg_sets", to convey what is in the data (api_kwargs) and the fact that there are multiple (sets).

pipeline/pipelines/batch_generation_pipeline.py
```diff
@@ -0,0 +1,133 @@
...
+        # the language set in _get_configs
+        del kwargs['language']
+
+        repo_root = kwargs['repo_root']
+        api_config_yamls = _get_api_configs(api_config_pattern, apis)
+
+        for api_kwargs in _get_configs(
+                api_config_yamls, languages, artman_language_yaml, repo_root):
+
+            api_kwargs.update(kwargs)
```
garrettjonesgoogle: I'm slightly concerned about one api x language combination setting a value that the next one doesn't set, but due to how this works, it doesn't get cleared out.

pipeline/pipelines/batch_generation_pipeline.py
michaelbausor: There is not a strong reason to do it here rather than lower down, just that we know the language at this point and don't need to set it for each API in the inner loop. I can move it down to api_kwargs

pipeline/pipelines/gapic_generation_pipeline.py
```diff
@@ -52,20 +53,33 @@ def __init__(self, **kwargs):
...
 
 
+class GapicClientBatchPipeline(batch_gen.BatchPipeline):
+
+    def __init__(self, **kwargs):
+        super(GapicClientBatchPipeline, self).__init__(
+            _get_batch_pipeline_tasks, **kwargs)
+
+
+def _get_batch_pipeline_tasks(**kwargs):
```
garrettjonesgoogle: This isn't really actionable, but I wanted to call out that this function is actually a factory factory.

pipeline/pipelines/batch_generation_pipeline.py
michaelbausor: This shouldn't happen with the way this is set up. Each api_kwargs is independent of previous ones, and then values are overwritten by what is in kwargs, so kwargs is never changed and no information (i.e. settings) should leak between one api x language combination and another.

garrettjonesgoogle: Ahh, I see now, sounds good.

pipeline/pipelines/batch_generation_pipeline.py
michaelbausor: What about "_get_api_kwarg_dicts"?

garrettjonesgoogle: Yeah that's better.

pipeline/pipelines/gapic_generation_pipeline.py
michaelbausor: Will rename to "_make_batch_pipeline_tasks"

pipeline/pipelines/core_generation_pipeline.py
```diff
@@ -55,13 +55,10 @@ class _GoCoreTaskFactory(code_gen.TaskFactoryBase):
...
-            protoc_tasks.ProtoCodeGenTask('CoreProtoGen',
-                                          inject=kwargs),
-            protoc_tasks.GoExtractImportBaseTask('ExtractGoPackageName',
-                                                 inject=kwargs),
-            protoc_tasks.GoLangUpdateImportsTask('UpdateImports',
-                                                 inject=kwargs)]
+        tasks = [protoc_tasks.ProtoCodeGenTask,
+                 protoc_tasks.GoExtractImportBaseTask,
+                 protoc_tasks.GoLangUpdateImportsTask]
+        return task_utils.instantiate_tasks(tasks, kwargs)
```
garrettjonesgoogle: We have a confusing situation where instantiate_tasks can be called from both base pipeline code and task factory code. I think it would be simpler if we dealt only with non-instantiated tasks for as long as possible, and confined task instantiation to one place (or as few places as possible).

michaelbausor: That is what I had originally. However, there is a problem when implementing this in the batch pipeline, because the get_tasks function for the BatchTaskFactory doesn't return only tasks - it returns flows which contains tasks, so these must necessarily be instantiated. While it would be possible to change BatchTaskFactory.get_tasks to return a long list of all the tasks, this would lose the structure of api x language tasks, and would prevent changing from a linear flow to an unordered flow in the future.

One thing that could be done in core_generation_pipeline is to add a base class that does the instantiation, like gapic_generation_pipeline and grpc_generation_pipeline

pipeline/pipelines/batch_generation_pipeline.py
michaelbausor: Yes, it would be reasonable to make it an error. This could also be added to the gapic config pipeline, which doesn't need the language.

Right now language defaults to None, which is why it is necessary to delete it - I could change this behaviour and add error checks in the batch pipeline and gapic config pipelines.

garrettjonesgoogle: Wait, would that mess up the ability to run the batch pipeline on only one language?

michaelbausor: Right now there is a `language` setting which is in the command line, and there is the `languages` setting which is in the batch config. So the `languages` setting will still be able to be set to just a single language.

Maybe `languages` should be renamed to `batch_languages`, or something less similar

pipeline/pipelines/core_generation_pipeline.py
garrettjonesgoogle: Another option is to use a recursive structure, where each element in a task list is either a task class or another task list, and the full tree could be instantiated before it needs to be run.

pipeline/pipelines/batch_generation_pipeline.py
garrettjonesgoogle: What I would really like is to be able to give 1 language and have it run the single pipeline, and if I gave multiple languages, have it run the batch pipeline.

pipeline/pipelines/core_generation_pipeline.py
michaelbausor: Then we would also need a way of indicating whether a list of flows/tasks needed to be performed serially (linear flow) or could be done in parallel (unordered flow) or, if it ever comes up, using something more complex like taskflow's graph flow (http://docs.openstack.org/developer/taskflow/patterns.html#module-taskflow.patterns.graph_flow)

garrettjonesgoogle: We could make it simple, and say that anything within a single api x language combination it is linear, but above that level, they are unordered.

michaelbausor: I guess I am not really seeing the benefits of this approach, or how it would work for our batch pipeline. Consider the situation after we have added a task that creates a PR after generating all of the api x language combinations. Our list might look like this:

`[PrepGoogleApisDir, [[Gapic1-java-pubsub, Gapic2-java-pubsub], [Gapic1-java-logging, Gapic2-java-logging]], CreatePR]`

So somehow we need to encode that the top-level list needs to be serial, and each of the individual `[Gapic1, Gapic2]` lists needs to be serial, but that `[Gapic1-java-pubsub, Gapic2-java-pubsub]` can be done in parallel with `[Gapic1-java-logging, Gapic2-java-logging]`. Then add in the possibility that we might want to make some individual task like `Gapic1` actually be a flow (for example, gapic publishing could be a flow) and we are basically recreating the existing taskflow structure, just to avoid calling the instantiation function.

pipeline/pipelines/batch_generation_pipeline.py
michaelbausor: I think this is something that should be done in the wrapper that @geigerj is working on, rather than in artman, because it breaks the current relationship between the pipeline command line argument and a particular pipeline being run.

I definitely agree that that is a good place to end up.

garrettjonesgoogle: I can live with that for now, although it seems like we could just use a single pipeline for both cases (the batch pipeline).

pipeline/pipelines/core_generation_pipeline.py
garrettjonesgoogle: ok, I guess my desire for simplification was a pipe dream here. Carry on...

garrettjonesgoogle:
LGTM once you can get Travis to pass.

PR #70, opened 2016-08-23
Fixing CoreProtoPipeline when --language is passed

garrettjonesgoogle:


execute_pipeline.py
```diff
@@ -145,6 +145,8 @@ def _load_config_spec(config_spec, config_sections, repl_vars):
         all_config_data = yaml.load(config_file)
     for section in config_sections:
         data.update(all_config_data[section])
+    if language in all_config_data:
```
michaelbausor: In the case that a list of config sections is specified on the commend line (after a ':' symbol) this will still try to load the language section (and will do so twice if it is listed).

garrettjonesgoogle: Is it a problem to always add the language? (Will someone have reason to leave it out for a file which contains a section for that language, if they are actually processing that language?) 

Also, is it a problem to load the language section twice? Since an update is being done here, it seems like the result would be idempotent. I can see the argument for avoiding the re-update for clarity, though.

michaelbausor: It is idempotent, so no behaviour problems.

I can't think of a case where someone would want to leave out the language section, but it seems desirable to make the command line args capable of fully specifying what you want to load.

garrettjonesgoogle: But conversely, say we add a special section, and the user wants to specify that, but also they don't want to have to include the language too because that would be redundant with the --language arg. 

I guess I would argue that if someone needs an exact list of sections and wants to exclude --language, maybe we should make special syntax for that when it actually comes up as a use case.

michaelbausor: Fair enough - let's leave it as is for now

michaelbausor:
LGTM

PR #71, opened 2016-08-23
Add requirements into seteup.py

jmuk:
pip install command looks up the requirements listed here.
Thus sometimes it will fail to install some dependencies when
the user installs artman through pip-install command.


bjwatson:
LGTM

PR #75, opened 2016-08-25
Improve error reporting for bad local_repo

### LINKED ISSUE #74: Remote execution fails when local_repo is too big
### Observed behavior

When the directory specified by the local_repo dir is too big, remote execution fails with the confusing message "taskflow.exceptions.StorageFailure: Storage backend internal error". This is a frequent error case if the directory pointed to is actually a git repo (i.e., also contains associated git metadata).
### Expected behavior

Artman should validate that the size of local_repo is < 1 MB.
### Action

Accumulate the size of local_repo here: https://github.com/googleapis/artman/blob/e83b08313c7fff2cb1f5ca6f716220050d1d4fd9/execute_pipeline.py#L221
and throw a meaningful exception if it is too big.

geigerj:
Previously, a confusing error occurred remotely if the local_repo was
too big. Instead, validate the size locally and raise a meaningful
exception.

Fixes #74

### DIFF
--- a/pipeline/tasks/io_tasks.py
+++ b/pipeline/tasks/io_tasks.py
@@ -87,8 +87,7 @@ class PrepareUploadDirTask(task_base.TaskBase):
             raise ValueError(
                 'Size of merged local_repo was {}, exceeding the limit of {} '
                 'bytes; reduce the size of local_repo'.format(
-                    size,
-                    _ZOOKEEPER_NODE_DATA_SIZE_LIMIT))
+                    size, _ZOOKEEPER_NODE_DATA_SIZE_LIMIT))
 
 
 class CleanupTempDirsTask(task_base.TaskBase):
--- a/pipeline/tasks/io_tasks.py
+++ b/pipeline/tasks/io_tasks.py
@@ -82,6 +82,13 @@ class PrepareUploadDirTask(task_base.TaskBase):
 
     def execute(self, repo_root, tarfile):
         self.exec_command(['tar', '-C', repo_root, '-zcvf', tarfile, '.'])
+        size = os.path.getsize(tarfile)
+        if size > _ZOOKEEPER_NODE_DATA_SIZE_LIMIT:
+            raise ValueError(
+                'Size of merged local_repo was {}, exceeding the limit of {} '
+                'bytes; reduce the size of local_repo'.format(
+                    size,
+                    _ZOOKEEPER_NODE_DATA_SIZE_LIMIT))
 
 
 class CleanupTempDirsTask(task_base.TaskBase):
@@ -120,18 +127,10 @@ class PrepareGoogleapisDirTask(task_base.TaskBase):
         repo_dir = os.path.join(repo_root, "googleapis")
         # Write/overwrite the additonal files into the repo_dir so that user
         # can include additional files which are not in the public repo.
-        total_bytes = 0
         for f, content in files_dict.iteritems():
-            total_bytes += os.path.getsize(f)
             filename = os.path.join(repo_dir, f)
             if not os.path.exists(os.path.dirname(filename)):
                 os.makedirs(os.path.dirname(filename))
             with open(filename, "w+") as text_file:
                 text_file.write(base64.b64decode(content))
-        if total_bytes > _ZOOKEEPER_NODE_DATA_SIZE_LIMIT:
-            raise ValueError(
-                'Size of merged local_repo was {}, exceeding the limit of {} '
-                'bytes; reduce the size of local_repo'.format(
-                    total_bytes,
-                    _ZOOKEEPER_NODE_DATA_SIZE_LIMIT))
         return repo_dir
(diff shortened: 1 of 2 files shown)


pipeline/tasks/io_tasks.py
```diff
@@ -24,6 +24,11 @@
 from pipeline.tasks import task_base
 
 
+# Maximum amount of data, in bytes, that can be stored in a zookeeper node. See
+# http://zookeeper.apache.org/doc/r3.1.2/api/org/apache/zookeeper/ZooKeeper.html
+_ZOOKEEPER_NODE_DATA_SIZE_LIMIT = 1048576
```
bjwatson: Define this as `1024 * 1024` or `2**20` so it's more clear that this is 1MB.

pipeline/tasks/io_tasks.py
```diff
@@ -115,10 +120,16 @@ def execute(self, repo_root, files_dict={}):
...
+            total_bytes += os.path.getsize(f)
             filename = os.path.join(repo_dir, f)
             if not os.path.exists(os.path.dirname(filename)):
                 os.makedirs(os.path.dirname(filename))
             with open(filename, "w+") as text_file:
                 text_file.write(base64.b64decode(content))
+        if total_bytes > _ZOOKEEPER_NODE_DATA_SIZE_LIMIT:
+            raise ValueError('size of merged local_repo exceeds limit of {} '
+                             'bytes; reduce the size of local_repo'.format(
+                                 _ZOOKEEPER_NODE_DATA_SIZE_LIMIT))
```
bjwatson: Also output the size of the "merged local_repo". It's helpful for users to know how much they overshot the limit.

pipeline/tasks/io_tasks.py
```diff
@@ -115,10 +120,16 @@ def execute(self, repo_root, files_dict={}):
...
         # can include additional files which are not in the public repo.
+        total_bytes = 0
         for f, content in files_dict.iteritems():
+            total_bytes += os.path.getsize(f)
             filename = os.path.join(repo_dir, f)
             if not os.path.exists(os.path.dirname(filename)):
                 os.makedirs(os.path.dirname(filename))
             with open(filename, "w+") as text_file:
                 text_file.write(base64.b64decode(content))
+        if total_bytes > _ZOOKEEPER_NODE_DATA_SIZE_LIMIT:
```
bjwatson: IIUC, this code is testing if just the custom files add up to more than 1MB (unzipped). My understanding is that the limit applies to the entire `googleapis/` repository _after_ it has been tarred and zipped.

pipeline/tasks/io_tasks.py
geigerj: Done

pipeline/tasks/io_tasks.py
geigerj: Done

pipeline/tasks/io_tasks.py
geigerj: I wasn't sure if the directory was stored archived or not in the zookeeper node and figured this was safer since it should usually be smaller than the unzipped file size if so. I will change it, though.

bjwatson: Also, should this calculation consider _all files_ in the merged googleapis, or just the custom files? My understanding is the former.

geigerj: The former is correct. The whole thing is uploaded after the merge.

pipeline/tasks/io_tasks.py
```diff
@@ -82,6 +82,13 @@ class PrepareUploadDirTask(task_base.TaskBase):
 
     def execute(self, repo_root, tarfile):
         self.exec_command(['tar', '-C', repo_root, '-zcvf', tarfile, '.'])
+        size = os.path.getsize(tarfile)
+        if size > _ZOOKEEPER_NODE_DATA_SIZE_LIMIT:
+            raise ValueError(
+                'Size of merged local_repo was {}, exceeding the limit of {} '
```
bjwatson: Size -> "Zipped size"

bjwatson: Also, use present tense: was -> is

bjwatson:
Since `codecov` is complaining, there's probably a relatively simple way to test this using two test cases:

1) Break out the validation code into its own function (i.e., `PrepareUploadDirTask._check_size(file)`). Generate a dummy "zip file" full of zeros that is exactly `_ZOOKEEPER_NODE_DATA_SIZE_LIMIT` in size, and run it through the `PrepareUploadDirTask._check_size()` code. Should succeed. 

2) Do the same with a "zip file" full of zeros that is exactly `_ZOOKEEPER_NODE_DATA_SIZE_LIMIT + 1` in size. Should raise exception.

geigerj:
Added a test. Codecov is happy now.

geigerj:
@shinfan @michaelbausor Adding some extra folks to this review in case @bjwatson needs to take ownership of this PR while I am OOO.

bjwatson:
LGTM

PR #76, opened 2016-08-25
Don't assume string exists during macro expansion

geigerj:
This allows artman to process incomplete configs as in
https://github.com/googleapis/googleapis/pull/104


bjwatson:
LGTM

PR #77, opened 2016-08-26
Fix various issues with running remote pipeline.

bjwatson:
This round was specifically focused on debugging GAPIC config generation
issues.


pipeline/tasks/gapic_tasks.py
```diff
@@ -61,10 +61,7 @@ def _move_to(self, gapic_config_dir, gapic_api_yaml):
         elif len(gapic_api_yaml) == 0:
             raise ValueError(error_fmt + 'No location specified')
         conf_out = os.path.abspath(gapic_api_yaml[0])
-        if os.path.exists(conf_out):
-            raise ValueError(error_fmt + 'File already exists')
```
bjwatson: This test seems to serve no purpose, since I can regenerate a GAPIC config if I want to. This was ultimately what was breaking GapicConfigPipeline for me.

geigerj: IIRC this is here because in local mode the pipeline automatically copies the output to the location of the GAPIC config specified in the artman config. That means that if you have any hand edits to an existing GAPIC config that you haven't `git commit`d, you will lose them.

@shinfan Can you confirm that?

USAGE.rst
```diff
@@ -39,26 +39,31 @@ To execute a build pipeline locally, try the following example command:
...
        --pipeline_kwargs={\'sleep_secs\':2} \
-       --env remote SamplePipeline
+       --env=<unique_jobboard_name> SamplePipeline
 
 Please note, there is no guarantee that your pipeline job will be claimed by the
-conductor running on your machine. It can also be picked up by other conductor
-processes. You can run the second command for multiple times, and chances are
-good that your conductor will pick up one job at least.
+conductor running on the first machine unless you choose a jobboard name that no one
+else is using. Otherwise, it can be picked up by other conductor processes.
```
geigerj: Should we reference the fact that there is a "special" remote environment always running in k8s if you don't want to start your own conductor?

pipeline/pipelines/grpc_generation_pipeline.py
```diff
@@ -97,6 +97,8 @@ def get_grpc_task_factory(language):
     cls = _GRPC_TASK_FACTORY_DICT.get(language)
     if cls:
         return cls()
+    elif language is None:
```
geigerj: note: typically you would just do this in the argparse configuration by making that flag required, but I actually prefer that you leave it this way because it will make transitioning execute pipeline from a script to an orchestrator easier when the artman wrapper is available.

pipeline/tasks/gapic_tasks.py
shinfan: Correct, a merge step here is probably necessary.

USAGE.rst
bjwatson: I thought about that, but it's not currently accessible outside Google, so it's not useful in this documentation.

pipeline/tasks/gapic_tasks.py
bjwatson: Is it really necessary? How often are all three of these conditions satisfied:

1) Running in local mode
2) Re-generating an existing GAPIC config (rather than a new one)
3) You have unstaged changes (i.e. not `git add`'d) in your googleapis sandbox

IIUC, that's the only time this would be a problem. Maybe we can just treat this as a user expectation issue?

pipeline/pipelines/grpc_generation_pipeline.py
bjwatson: Good point, but I'll leave this alone as you say.

pipeline/tasks/gapic_tasks.py
garrettjonesgoogle: I was just doing this the other day when using config gen to auto-generate the new IAM methods and copy them onto the existing gapic yaml. Admittedly, I had to hack the gapic yaml path so that it would generate to a different file path than the one that is checked in, so that it wouldn't get overwritten.

geigerj: Honestly, I would prefer that artman just didn't try to put files directly into `googleapis`.

I do frequently edit the GAPIC config when I'm starting on a new API, since there's still significant manual work required sometimes, and I don't always remember `git add` it right after I make my edits. I then run the `GapicClient` artman pipeline -- I can imagine a frustrating situation where I accidentally run the `GapicConfig` pipeline instead and lose all my changes.

USAGE.rst
geigerj: Ah, that's true.

pipeline/tasks/gapic_tasks.py
bjwatson: @garrettjonesgoogle Was condition #3 satisfied when you were doing that?

@geigerj That could also be argued as a command naming issue. I used to alias `git commit` as `git ci`, and `git checkout` as `git co`. Then I realized that running `git co .` in a sandbox full of unstaged changes when I wanted to run `git ci .` could really ruin my day. I changed the checkout alias to `git cho`, which is different enough that I'm no longer worried about it.

garrettjonesgoogle: Yes, I had unstaged changes. It would have made me unhappy if I had accidentally changed artman_pubsub.yaml back to the old path and regenerated on top of my edits. 

You silly people and your aliases. Just do "ctrl+shift+r" and type "comm" for a reverse search through your command-line history, and "git commit" will pop up; likewise for "check" to get "git checkout", or a substring of any command for that matter. Having the full commands in your history is easier to understand and more portable.

bjwatson: Everyone has their own way of working. I use Ctrl-R for infrequently used commands in my history, but once they become bread-and-butter, I create aliases for them. They have less cognitive overhead than making sure I have the correct variant of a command from my history.

I'll drop this change from this PR, since I think it's the only controversial bit. I'll add a TODO comment that perhaps it should only be done for local mode, due to the behavior of overwriting an active sandbox.

bjwatson: https://github.com/googleapis/artman/issues/80

bjwatson:
@garrettjonesgoogle PTAL

USAGE.rst
```diff
@@ -39,26 +39,31 @@ To execute a build pipeline locally, try the following example command:
      python execute_pipeline.py \
        --pipeline_kwargs={\'sleep_secs\':2} SamplePipeline
 
-To execute a build pipeline remotely, start two consoles and run the following command
-in one console:
+Remote
+******
+
+To execute a build pipeline remotely, start two consoles (on the same or different machines) and run 
+the following command in one console:
```
garrettjonesgoogle: I know this isn't integral to this PR, but I'm confused about this - is this to run the pipeline on the console running the conductor, or to run the pipeline on a third machine in the Google infrastructure?

USAGE.rst
```diff
@@ -39,26 +39,31 @@ To execute a build pipeline locally, try the following example command:
...
+       --env=<unique_jobboard_name> SamplePipeline
 
 Please note, there is no guarantee that your pipeline job will be claimed by the
-conductor running on your machine. It can also be picked up by other conductor
-processes. You can run the second command for multiple times, and chances are
-good that your conductor will pick up one job at least.
+conductor running on the first machine unless you choose a jobboard name that no one
+else is using. Otherwise, it can be picked up by other conductor processes.
+
+Be sure you are running Java 7. There might be issues with Java 8.
```
garrettjonesgoogle: This doesn't quite make sense. You run programs with Java 7 or 8, you don't just "run Java 7" and let it loose...

USAGE.rst
garrettjonesgoogle: Ok I see from Jacob's comment what's going on. These instructions are somewhat unclear about what console is doing what; it seems to be that the first console is the "remote" console.

USAGE.rst
bjwatson: Right, I was referring to which version of Java is in the PATH. I was seeing an issue when that was version 8 on my workstation, and I fixed that issue by switching to 7. Unfortunately, I can't reproduce the issue now, so I'm dropping this statement until someone sees it again.

bjwatson:
@garrettjonesgoogle I clarified that USAGE.rst text. PTAL

garrettjonesgoogle:
LGTM

PR #78, opened 2016-08-26
Fix a typo (missing providing values) in a Ruby Gapic pipeline.

jmuk:


pipeline/tasks/gapic_tasks.py
```diff
@@ -134,7 +134,7 @@ def execute(self, language, api_name, final_repo_dir):
         # Such APIs don't need packman.
         if ((language == 'ruby' or language == 'nodejs') and
                 task_utils.is_output_gcloud(final_repo_dir)):
-            return
+            return final_repo_dir
```
geigerj: This logic doesn't make sense to me -- if `package_dir == final_repo_dir` always, why do we need two variables?

geigerj: Ah nevermind -- I see. It is different in `protoc_tasks`

geigerj: But this code can be simpler by having a single `return final_repo_dir` outside the conditional.

jmuk: Done

geigerj:
LGTM after lint errors fixed.

PR #83, opened 2016-08-30
Deleting JavaCorePackmanTask, supporting dependencies

garrettjonesgoogle:
This depends on https://github.com/googleapis/packman/pull/99 , and depends on an update in googleapis to include the new dependencies field.


garrettjonesgoogle:
Corresponding change in googleapis: https://github.com/googleapis/googleapis/pull/117

pipeline/tasks/protoc_tasks.py
```diff
@@ -359,24 +359,18 @@ def execute(self, language, api_name, output_dir, src_proto_path,
...
 
     def execute(self, language, api_name, output_dir, src_proto_path,
-                import_proto_path, packman_flags=None, repo_dir=None):
-        packman_flags = packman_flags or ['--experimental_alt_java']
+                import_proto_path, packman_flags=[], repo_dir=None,
+                dependencies=[]):
+        if len(packman_flags) == 0:
+            packman_flags.append('--experimental_alt_java')
+            for dep in dependencies:
+                packman_flags.append('--dependencies')
```
michaelbausor: Maybe the packman flag should be singular `--dependency` if it is repeated for each additional dependency

garrettjonesgoogle: You make sense. I can change it.

garrettjonesgoogle:
I renamed the flag to proto_gen_pkg_dep, because there was already a notion of 'dependencies' in packman (dependency versions like grpc, protobuf, etc).

michaelbausor:
LGTM pending packman change

garrettjonesgoogle:
I made a rename because I think 'dependencies' was too general - PTAL

michaelbausor:
What about `proto_dependencies`?

garrettjonesgoogle:
That implies dependencies on proto files, as opposed to generated packages.

jmuk:
LGTM -- I have no opinions on the flag names. `proto_gen_pkg_deps` would be good.

michaelbausor:
Fair enough, LGTM

garrettjonesgoogle:
*ok*, I had to make yet another change here. It turned out that removing iam/v1 from src_proto_path broke toolkit generation, so I had to add another attribute, desc_proto_path, that gets added when ProtoDescGenTask is run, but not when packman is run. This will also require an update in googleapis in the other PR.

garrettjonesgoogle:
Sneaking in another tiny update... it's ready now, I promise...

test/fake-repos/fake-mixin-proto/fake_mixin.proto
```diff
@@ -0,0 +1,8 @@
+// This is a fake proto file for testing.
+
+syntax = "proto3";
+
+package google.fake_mixin
```
michaelbausor: Is it necessary/useful to put some content into this proto, and have library.proto use it as a mixin?

pipeline/tasks/protoc_tasks.py
```diff
@@ -206,20 +206,22 @@ class ProtoDescGenTask(task_base.TaskBase):
     """Generates proto descriptor set"""
     default_provides = 'descriptor_set'
 
-    def execute(self, src_proto_path, import_proto_path, output_dir,
-                api_name, toolkit_path):
-        print 'Compiling descriptors {0}'.format(
-            _find_protos(src_proto_path))
+    def execute(self, src_proto_path, desc_proto_path, import_proto_path,
```
jmuk: `desc_proto_path` could be an optional parameter (move at the end with `desc_proto_path = None` notation, and then initialize with an empty list when `None`). In that case that is not a `GAPIC_REQUIRED` parameter, but it would simplify the artman config a bit, because, anyways, `desc_proto_path` would be empty in many APIs.

test/fake-repos/fake-mixin-proto/fake_mixin.proto
garrettjonesgoogle: I don't think so. It's really only the presence of this file that matters from the artman perspective. We test the mixing-in in toolkit.

pipeline/tasks/protoc_tasks.py
garrettjonesgoogle: Ahh ok, that sounds better.

garrettjonesgoogle:
Ok, made desc_proto_path optional, PTAL

pipeline/tasks/protoc_tasks.py
```diff
@@ -359,24 +361,18 @@ def execute(self, language, api_name, output_dir, src_proto_path,
 class JavaGrpcPackmanTask(GrpcPackmanTask):
 
     def execute(self, language, api_name, output_dir, src_proto_path,
-                import_proto_path, packman_flags=None, repo_dir=None):
-        packman_flags = packman_flags or ['--experimental_alt_java']
+                import_proto_path, packman_flags=[], repo_dir=None,
+                proto_gen_pkg_deps=[]):
+        if len(packman_flags) == 0:
+            packman_flags.append('--experimental_alt_java')
```
jmuk: This can be problematic. Consider specifying `packman_flags=None`, and initialize as `packman_flags = [] if not packman_flags` at the beginning of the method.

In Python, the default value will be instantiated once and reused. And append/extend on a list is a destructive operations, so the default value itself (for further invocations) will be updated. For example:
```python
def foo(x, y=[]):
    y.append(x)
    return y

foo(1)  # => [1]
foo(2)  # => [1, 2]
```
See also: http://docs.python-guide.org/en/latest/writing/gotchas/

garrettjonesgoogle: oh jeez... I will fix it.

garrettjonesgoogle:
defaulting issue fixed, PTAL

jmuk:
LGTM

michaelbausor:
LGTM also. Very interesting link Jun!

PR #84, opened 2016-08-31
Added a task to RubyGrpcPipeline to copy files to final repo dir.

### LINKED ISSUE #72: Stop using packman for Ruby gRPC client generation
Derived from #69:

Currently Ruby gRPC client pipeline use packman to generate gRPC files. However, that does not fit with the design of gcloud-ruby, where a package contains all of hand-written layers, Gapic-generated files, and gRPC-generated files.
That is Cloud APIs only, but there are no strong reasons to split packages for Gapic files and gRPC files in Ruby.

Because of that, we should stop using packman for Ruby's client pipeline. Instread, similar to Java, we should use protoc_tasks we already have.

Also, from https://github.com/googleapis/artman/issues/69#issuecomment-242172115,

> Currently we get protoc from linuxbrew here: https://github.com/grpc/homebrew-grpc
> That site indicates that the linuxbrew method is deprecated for all languages except C/C++. 

Therefore, Ruby will use grpc-tools rubygem (https://rubygems.org/gems/grpc-tools).

Cc: @bjwatson @geigerj

Landrito:
Resolves #72.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -69,6 +69,7 @@ RUN npm install -g googleapis-packman@0.8.0
 RUN gem install rubocop --version '= 0.39.0' --no-ri --no-rdoc
 RUN gem install bundler --version '= 1.12.1' --no-ri --no-rdoc
 RUN gem install rake --version '= 10.5.0' --no-ri --no-rdoc
+RUN gem install grpc-tools --version '=1.0.0' --no-ri --no-rdoc
 
 # Install couple of git repos
 WORKDIR /
--- a/Dockerfile
+++ b/Dockerfile
@@ -69,7 +69,6 @@ RUN npm install -g googleapis-packman@0.8.0
 RUN gem install rubocop --version '= 0.39.0' --no-ri --no-rdoc
 RUN gem install bundler --version '= 1.12.1' --no-ri --no-rdoc
 RUN gem install rake --version '= 10.5.0' --no-ri --no-rdoc
-RUN gem install grpc-tools --version '=1.0.0' --no-ri --no-rdoc
 
 # Install couple of git repos
 WORKDIR /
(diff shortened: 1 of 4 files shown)


jmuk:
It's sad, but @garrettjonesgoogle pointed out that possibly it's better to keep packman task for protoc. That's actually more consistent with other language pipelines. Sorry for my misunderstanding.

Let's revive the packman tasks, but
- add --proto_compiler flag and --proto_compiler_args flag (if necessary) to the packman task. IIRC Python one is doing so already.
- then, your copying task can be used to copy the result of packman's proto compilation to the expected file location.
- also -- please fix Dockerfile to install grpc_tools_ruby_protoc command.

Landrito:
@jmuk, PTAL.

The packman task now uses the grpc-tools proto compiler as well as copies the files to the main repo.

jmuk:
Sorry I'm changing opinions so frequently, but as I commented at https://github.com/googleapis/packman/pull/104, I got unsure about the approach of `--proto_compiler` approach. I now think that the logic of `grpc_tools_ruby_protoc` if `language == ruby` should be in packman instead.
So sorry, but with that RubyGrpcPackmanTask wouldn't be necessary. Let's talk about this tomorrow.

By the way -- it's weird googlebot warns about CLA. Let's also talk about that.

bjwatson:
@Landrito Looks like you haven't joined the `google` and `googleapis` groups. That's why you're getting the CLA request.

bjwatson:
I added a relevant comment to https://github.com/googleapis/packman/pull/104 (the Python analogue) for this discussion.

Landrito:
Ah I see. I got added to the googleapis group but not the google group. It should be fixed now. Thanks for the heads up.

jmuk:
Hey -- you don't have to close this issue. Are you trying to create a new PR?

The difference of Python and Ruby is that currently Ruby is working. We just want to switch to the right tool.
But, the copy task will be still a good thing to do.

bjwatson:
BTW, the correct Python analogue is actually https://github.com/googleapis/artman/pull/86

Landrito:
@jmuk This PR has been reduce down to only have the copy task. PTAL

jmuk:
LGTM!
Sorry for the confusion along with my saying.

Landrito:
No worries! Thanks for the review.

PR #85, opened 2016-08-31
Add batch output dir name setting

michaelbausor:


michaelbausor:
See also https://github.com/googleapis/googleapis/pull/96

test/testdata/googleapis_test/gapic/api/artman_library.yaml
```diff
@@ -12,15 +12,22 @@ common:
   enable_batch_generation: True
 java:
   final_repo_dir: ${REPOROOT}/test/testdata/test_output/final
+  batch_dir_name: java
```
garrettjonesgoogle: will batch_dir_name ever be different from the language?

michaelbausor: Yes, see the googleapis PR (https://github.com/googleapis/googleapis/pull/96). For php, it will be `gcloud-php`, while for java it will be `gcloud-java/gcloud-java-<API>`

PR #88, opened 2016-09-07
Update pip, packman in dockerfile

geigerj:
Otherwise, installation of grpcio-tools fails if docker pulls in
an older version of pip. (The installation requires pip>=8)

For packman, new version is needed for flags required by
artman.


bjwatson:
LGTM

PR #90, opened 2016-09-07
Fix download condition

### LINKED ISSUE #89: Remote mode fails to download output
### Expected behavior
After a pipeline completes successfully in remote mode, the results are downloaded as a zip file.

### Observed behavior
Artman does not attempt to download result.

### Possible cause
[A recent commit](https://github.com/googleapis/artman/commit/05d74055a85d6295887a76709da17dc648e019c8#diff-41ef7726e048e1a4cd48228eeada3686R48) changed the names of the tasks, but the [conditional that triggers the download](https://github.com/googleapis/artman/blob/master/execute_pipeline.py#L69) still expects the old name. Since there is no task by the old name, the download is never triggered.

/cc @michaelbausor @bjwatson

geigerj:
Previously download was only triggered by a certain task name which
no longer exists.

Fixes #89

### DIFF
--- a/execute_pipeline.py
+++ b/execute_pipeline.py
@@ -66,7 +66,7 @@ def main(args):
         task_details, flow_detail = job_util.fetch_job_status(jb, env)
 
         for task_detail in task_details:
-            if task_detail.name == 'BlobUploadTask' and task_detail.results:
+            if task_detail.name.startswith('BlobUploadTask') and task_detail.results:
                 bucket_name, path, _ = task_detail.results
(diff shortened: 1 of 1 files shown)


bjwatson:
I don't understand this PR. I don't see any other tasks that start with 'BlobUploadTask'.

geigerj:
Full context is in #89. Basically a previous commit added some additional text to the task names, which means string equality doesn't work as a condition.

bjwatson:
Ok, I see. LGTM

PR #92, opened 2016-09-08
Switch to installing packman from source

geigerj:
This means we don't need to re-release packman and update the
Dockerfile every time we roll out an updated configuration of the
artman remote.

Note that eventually when artman is versioned we'll want to switch
back to pegging packman.


jmuk:
LGTM

PR #94, opened 2016-09-13
Remove THISDIR macro expansion

### LINKED ISSUE #93: Deprecate ${THISDIR} expansion
After https://github.com/googleapis/googleapis/pull/130 there will be no remaining use for the `THISDIR` macro, which is incompatible with the remote mode for artman. We can remove the expansion logic altogether.

geigerj:
- THISDIR is no longer used in the real googleapis
- THISDIR causes problems when running in remote mode
- [geigerj/artman wrapper branch](https://github.com/geigerj/artman/tree/wrapper) relies on this change, but it's not directly related to the wrapper, so it seems good to separate it into its own PR.

Fixes #93

### DIFF
--- a/pipeline/utils/config_util.py
+++ b/pipeline/utils/config_util.py
@@ -16,7 +16,6 @@
 """Utils related to config files"""
 
 
-import os
 import yaml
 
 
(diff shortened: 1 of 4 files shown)


geigerj:
@bjwatson @michaelbausor PTAL when you have a chance

ethanbao:
LGTM

PR #96, opened 2016-09-14
Pass skip_packman as a pipeline_arg

### LINKED ISSUE #95: Add option to skip packman task
For Ruby and NodeJS APIs, due to the effort to integrate with the existing hand-written veneer, some artifacts such as package meta data (README.md, package.json, or gemspec) are not needed, so we can skip the packman task when generating the Gapic artifacts. However, for other APIs such as ErrorReporting, since there is no existing hand-written veneer code, these package meta data still need to be generated.

geigerj:
Formerly, packman was skipped for all Ruby/NodeJS Cloud GAPIC packages.
Actually, some packages require packman; the correct condtion is that
packman should be skipped if there is a handwritten layer on top of
GAPIC. This should be configured per-API instead of hard-coded into
artman.

Fixes #95

### DIFF
--- a/pipeline/tasks/gapic_tasks.py
+++ b/pipeline/tasks/gapic_tasks.py
@@ -130,11 +130,8 @@ class GapicCopyTask(task_base.TaskBase):
 class GapicPackmanTask(packman_tasks.PackmanTaskBase):
     default_provides = 'package_dir'
 
-    def execute(self, language, api_name, final_repo_dir):
-        # Some APIs will be a part of gcloud project for Ruby and NodeJS.
-        # Such APIs don't need packman.
-        if ((language != 'ruby' and language != 'nodejs') or
-                not task_utils.is_output_gcloud(language, final_repo_dir)):
+    def execute(self, language, api_name, final_repo_dir, skip_packman=False):
+        if not skip_packman:
             # TODO: Use TaskBase.exec_command()
             self.run_packman(language,
                              task_utils.packman_api_name(api_name),


geigerj:
See https://github.com/googleapis/googleapis/pull/135 for example configuration.

execute_pipeline.py
```diff
@@ -177,6 +177,7 @@ def _parse_args(args):
 
     cmd_args = ast.literal_eval(flags.pipeline_kwargs)
     pipeline_args.update(cmd_args)
+    pipeline_args['skip_packman'] = pipeline_args.get('skip_packman', False)
```
jmuk: You don't have to do this. Simply uses a default parameter in gapic_tasks.py, as `skip_packman=False`. Then pipeline won't raise any errors.

pipeline/tasks/gapic_tasks.py
```diff
@@ -130,11 +130,10 @@ def execute(self, final_repo_dir, gapic_code_dir):
 class GapicPackmanTask(packman_tasks.PackmanTaskBase):
     default_provides = 'package_dir'
 
-    def execute(self, language, api_name, final_repo_dir):
+    def execute(self, language, api_name, final_repo_dir, skip_packman=False):
         # Some APIs will be a part of gcloud project for Ruby and NodeJS.
         # Such APIs don't need packman.
```
jmuk: Wait a min, might be good to edit these comment lines.

geigerj: Removed

geigerj: (I could have edited to mention that this depends on presence/absence of handwritten layers, but it seems unhelpful to document Ruby/NodeJS-specific procedure in a generic artman function.)

PR #101, opened 2016-09-22
Upgrade gcloud dependency to google-cloud 0.19.x

bjwatson:
Fixes https://github.com/googleapis/artman/issues/103 and https://github.com/googleapis/artman/issues/102.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,9 @@
 oslo.utils>=3.4.0 # Apache-2.0
 pyyaml>=3.11
-taskflow==1.25.0
+taskflow>=1.25.0,<2.0.0
 kazoo>=2.2.1
 google-apitools
-gcloud==0.15.0
+google-cloud>=0.19.0
 grpcio-tools>=1.0.0
 yapf>=0.6.2 # Apache-2.0
 requests>=2.10.0,<3.0.0
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,9 +11,9 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     build-essential \
     man-db \
     pkg-config \
-    # Needed by gcloud-python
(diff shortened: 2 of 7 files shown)


requirements.txt
```diff
@@ -1,9 +1,9 @@
 oslo.utils>=3.4.0 # Apache-2.0
 pyyaml>=3.11
-taskflow==1.25.0
+taskflow>=1.25.0,<2.0.0
```
bjwatson: I'm just fixing this to be consistent with `setup.py`

bjwatson:
Never mind. This is more involved than I thought. I reverted the offending change from master, so the urgency is somewhat reduced.

There's still a version conflict in the self-service instructions that originally motivated what I was doing here.

shinfan:
Travis seems cannot import "google.cloud". Is this module name incorrect?

geigerj:
@shinfan what's especially strange is that the import error only happens in the `pylint` environment, not `pep8` or `py27`...

geigerj:
@shinfan From the Travis log:

```
pylint-errors installed:...
...google-cloud==0.19.0...
```

So `google-cloud` should be installed available in the `pylint-errors` environment. And

```
$ virtualenv venv
$ . venv/bin/activate
(venv) $ pip install google-cloud==0.19.0
  ...
(venv) $ python
Python 2.7.12 (default, Aug 24 2016, 08:19:55) 
[GCC 4.8.4] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import google.cloud
>>>
```

So the module `google.cloud` name seems OK.

shinfan:
Umm this is strange...Can you reproduce the issue locally?

lukesneeringer:
This is _way_ out of date at this point. Declaring it dead.

PR #107, opened 2016-09-30
Update USAGE for setting language and API in batch command

michaelbausor:


USAGE.rst
```diff
@@ -252,6 +252,32 @@ staging repository.
...
+batch_apis setting. For example, to run the batch command only Java and
+only for the PubSub API:
+
+  ::
+
+     python execute_pipeline.py \
+        --config "../googleapis/gapic/batch/common.yaml" \
+        --stage_output \
+        --pipeline_kwargs "{'batch_languages':'java','batch_apis':'pubsub'}" \
+        GapicClientBatchPipeline
```
garrettjonesgoogle: This is no longer the batch command, it is equivalent to GapicClientPipeline :-P Maybe switch the example to batch_apis only, without the language specified?

michaelbausor: Fair enough :-) Updated

garrettjonesgoogle:
LGTM

PR #108, opened 2016-10-05
Made artman not depend on packman for ruby grpc generation.

### LINKED ISSUE #72: Stop using packman for Ruby gRPC client generation

landrito:
Resolves #72.

### DIFF
--- a/pipeline/tasks/protoc_tasks.py
+++ b/pipeline/tasks/protoc_tasks.py
@@ -47,6 +47,7 @@ class _SimpleProtoParams:
     def grpc_out_param(self, output_dir):
         return '--grpc_out=' + self.code_root(output_dir)
 
+    @property
     def proto_compiler(self):
         return 'protoc'
 
@@ -72,6 +73,7 @@ class _JavaProtoParams:
     def grpc_out_param(self, output_dir):
         return '--grpc_out=' + self.code_root(output_dir)
 
+    @property
     def proto_compiler(self):
         return 'protoc'
 
@@ -100,6 +102,7 @@ class _GoProtoParams:
         # returns None.
         return None
 
+    @property
     def proto_compiler(self):
         return 'protoc'
 
@@ -124,6 +127,7 @@ class _PhpProtoParams:
     def grpc_out_param(self, output_dir):
         return '--grpc_out=' + self.code_root(output_dir)
 
+    @property
     def proto_compiler(self):
         return 'protoc'
 
@@ -149,6 +153,7 @@ class _RubyProtoParams:
     def grpc_out_param(self, output_dir):
         return '--grpc_out=' + self.code_root(output_dir)
 
+    @property
     def proto_compiler(self):
         return 'grpc_tools_ruby_protoc'
 
@@ -200,12 +205,10 @@ def _group_by_dirname(protos):
 
 
 def _protoc_header_params(proto_path,
-                          toolkit_path,
-                          proto_compiler='protoc'):
+                          toolkit_path):
     proto_path = proto_path[:]
     proto_path.append(_find_protobuf_path(toolkit_path))
-    return ([proto_compiler] +
-            ['--proto_path=' + path for path in proto_path])
+    return (['--proto_path=' + path for path in proto_path])
 
 
 def _protoc_desc_params(output_dir, desc_out_file):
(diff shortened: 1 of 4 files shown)


pipeline/tasks/protoc_tasks.py
```diff
@@ -124,6 +124,14 @@ def grpc_out_param(self, output_dir):
...
 }
 
+_PROTO_COMPILERS_MAP = {
+    'ruby': 'grpc_tools_ruby_protoc',
+    'java': 'protoc',
+    'go': 'protoc',
+    'csharp': 'protoc',
+    'php': 'protoc',
+}
+
```
jmuk: I think this should be merged into ProtoParams classes.

landrito: Done.

pipeline/tasks/protoc_tasks.py
```diff
@@ -69,6 +72,9 @@ def grpc_plugin_path(self, toolkit_path):
     def grpc_out_param(self, output_dir):
         return '--grpc_out=' + self.code_root(output_dir)
 
+    def proto_compiler(self):
+        return 'protoc'
```
garrettjonesgoogle: Because I prefer not to use inheritance in general...

pipeline/tasks/protoc_tasks.py
```diff
@@ -47,6 +47,9 @@ def grpc_plugin_path(self, dummy_toolkit_path):
...
```
landrito: Done.

pipeline/tasks/protoc_tasks.py
landrito: Noted. I'm going to keep the current architecture.

pipeline/tasks/protoc_tasks.py
```diff
@@ -163,10 +200,11 @@ def _group_by_dirname(protos):
 
 
 def _protoc_header_params(proto_path,
-                          toolkit_path):
+                          toolkit_path,
+                          proto_compiler='protoc'):
     proto_path = proto_path[:]
     proto_path.append(_find_protobuf_path(toolkit_path))
-    return (['protoc'] +
+    return ([proto_compiler] +
```
landrito: Done.

pipeline/tasks/protoc_tasks.py
```diff
@@ -218,8 +259,8 @@ def execute(self, src_proto_path, import_proto_path, output_dir,
         #   - it doesn't have to
         #   - and multiple invocation will overwrite the desc_out_file
         self.exec_command(
-            _protoc_header_params(
-                header_proto_path, toolkit_path) +
+            ['protoc'] +
```
jmuk: I think using `proto_params.proto_compiler` is fine here as well and I believe that's more consistent. You need to change a baseline for ruby_gapic_client_pipeline.

landrito: Since this task is being run in the gapic config pipeline having a proto params object doesn't make much sense since the proto params object is retrieved based on the language.

jmuk: Think about users who are only interested in Ruby (or think that remote environment which only takes care of Ruby). Considering those environment, requiring both `protoc` and `grpc_tools_ruby_protoc` doesn't sound great, because the latter has the same feature already.

landrito:
This should be ready for another pass per our chat offline.

PR #115, opened 2016-11-23
Refactor Python gRPC pipeline 

geigerj:
- Introduce a task that renames Python gRPC package declarations
- Use grpcio-tools directly instead of Packman to run codegen
- Move Python gRPC from Packman to MetadataGen

Note that the renaming task does not play nicely with the baseline tests, which do not permit the filesystem to be mocked, so I had to mock a portion of that task in the baseline test. Full coverage is provided through separate unit tests.

Updates #97 

cc: @jmuk @garrettjonesgoogle

PR #117, opened 2016-11-30
Turn gRPC codegen back on for Python

geigerj:
It was accidentally removed in #115.

Note that this change removes an flag that set the protoc gRPC plugin
to an empty value for Ruby and C#; presumably removing the empty flag
will have no effect.

FYI @swcloud @chrisdunelm regarding the modification to the Ruby/C# gRPC codegen command; see the baseline file change for details.


garrettjonesgoogle:
LGTM

Issue #116, opened 2016-11-29
API naming variable changes


geigerj:
@garrettjonesgoogle @bjwatson @omaray 

### Background

Some recent changes have introduced inconsistencies in how we refer to the gRPC Python packages -- just wanted to get input on how we should resolve this. Renaming the packages at a future date is a breaking change, and it's linked to some changes I want to make around how we treat the concept of "API names" in our tooling in connection to googleapis/toolkit#270, so I wanted to get some feedback before I implement this.

### Currently

- Artman currently has a single "API name" concept, which is configurable in the artman config.
- A sample API name might be google-cloud-logging-v2
- Our packaging tools (Packman, and its successor MetadataGen) further introduce two more API name variables:
    - The "version" is the last word in the API name, e.g., "v2"
    - The "short name" is the second-to-last word in the API name, e.g., "logging"
    - The "title name" is `Google {titlecased short name}`, e.g., "Google Logging". This is used in the autogenerated README to refer to the service.
- The gRPC Python package name is either `grpc-google-{shortname}-{version}` or `grpc-{api name}`, depending on where you look (i.e., either `grpc-google-logging-v2` or `grpc-google-cloud-logging-v2`). That's a problem.

### Proposal

- I want to get rid of the "API name" configuration value in artman. The existing variables will be replaced as follows:
    - Version --> independent configuration value in artman config
    - Short name --> independent configuration value in artman config
    - Title name --> use `title` attribute from service config (example)
    - API Name --> `google-cloud-{short name}-{version}`
- The Python package names will then be `grpc-{API Name}` and `gapic-{API Name}`.

Any objection or concerns with this?

geigerj:
See this commit for an example of how this might be done in artman naively:
  https://github.com/geigerj/artman/commit/b2ed7090962f6274186e76a6909e747c2cc07a1c

geigerj:
My concern with this is that it means the `artman` output directory *always* references Google Cloud, even if the API is not a Cloud API. Maybe we can introduce a new flag `cloud_api` to the artman `common.yaml` config file that only prefixes `google-cloud-` to the API name if the flag is set.

jskeet:
I wouldn't be at all surprised if the C# title needed to be specified by hand either way, to be honest. It might be worth having a "default" based on the four variables you've said, but have an escape hatch for any that are awkward for whatever reason.

geigerj:
[...]

PR #118, opened 2016-11-30
Fix typo in plugin param

geigerj:
#116 actually removed the flag name from the grpc plugin param. ~~I didn't notice this because no language is currently passing a grpc plugin to protoc directly.~~

EDIT: this fixes a breakage in the Ruby gRPC pipeline. #119 filed for why the baseline tests don't reflect that.


geigerj:
@jmuk @landrito I checked, and removing the `plugin` also solves the problem that @swcloud was seeing. Filed as #120

PR #122, opened 2016-12-02
Treat google.iam as common proto package

geigerj:
Context is the same as with https://github.com/googleapis/googleapis/pull/209; this is needed to prevent Python codegen from changing the package to `google.cloud.grpc.iam.[...]`.

### DIFF
--- a/pipeline/tasks/python_grpc_tasks.py
+++ b/pipeline/tasks/python_grpc_tasks.py
@@ -56,9 +56,12 @@ class PythonChangePackageTask(task_base.TaskBase):
         with open(common_protos_yaml) as common_protos_file:
             common_protos_data = yaml.load(common_protos_file)
 
-        # Treat google.protobuf as a common proto package, even though it is
-        # not included in the common-protos we generate.
-        common_protos = ['google.protobuf']
+        # Treat google.protobuf, google.iam as a common proto package, even
+        # though they are not included in the common-protos we generate.
+        #
+        # TODO (geigerj): remove 'google.iam' when it is included in the common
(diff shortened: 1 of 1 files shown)


pipeline/tasks/python_grpc_tasks.py
```diff
@@ -56,9 +56,12 @@ def execute(self, src_proto_path, import_proto_path, common_protos_yaml):
...
 
-        # Treat google.protobuf as a common proto package, even though it is
-        # not included in the common-protos we generate.
-        common_protos = ['google.protobuf']
+        # Treat google.protobuf, google.iam as a common proto package, even
+        # though they are not included in the common-protos we generate.
+        #
+        # TODO (geigerj): remove 'google.iam' when it is included in the common
+        # protos package.
+        common_protos = ['google.protobuf', 'google.iam']
```
bjwatson: Shouldn't this include everything in https://github.com/geigerj/toolkit/blob/59e0385f4dd580610a130cd57116ed6dc65ca0a9/src/main/java/com/google/api/codegen/py/PythonImportHandler.java#L42-L49?

geigerj: No -- `artman` has knowledge of the common protos *package* and which proto packages are contained within via https://github.com/googleapis/googleapis/blob/master/gapic/packaging/common_protos.yaml; `toolkit` doesn't know about this config file. There is a TODO in the `toolkit` code to fix this.

bjwatson: Ok, got it.

PR #123, opened 2016-12-09
Change naming configuration

geigerj:
Updates #116 

Currently api_name is parsed to derive many smaller pieces of information:
- a one-word short name
- a major version
- a path to the source protos in the googleapis repo for package metadata
- the name of the package

Instead of parsing this information out, it makes more sense to configure these values separately. This avoids corner cases where the short name may need to consist of multiple words (e.g., error-reporting, iam-admin), or where the name of the package may not match the path to the protos in the googleapis repo.

This PR separates out the short_name and version configuration. A subsequent PR will replace the source to the protos currently parsed from the api_name with the existing src_proto_path configuration value.


bjwatson:
No problems jump out at me, but I'd like someone else to review.

pipeline/tasks/gapic_tasks.py
```diff
@@ -82,9 +82,10 @@ class GapicCodeGenTask(task_base.TaskBase):
     default_provides = 'gapic_code_dir'
 
     def execute(self, language, toolkit_path, descriptor_set, service_yaml,
-                gapic_api_yaml, gapic_language_yaml, output_dir, api_name):
+                gapic_api_yaml, gapic_language_yaml, output_dir, short_name,
+                version):
         code_root = os.path.join(output_dir,
-                                 api_name + '-gapic-gen-' + language)
+                                 short_name + '-gapic-gen-' + language)
```
garrettjonesgoogle: Removing the version out of this path means that you can't build multiple major versions of the same API in one run.

pipeline/utils/task_utils.py
```diff
@@ -36,6 +36,13 @@ def gradle_task(toolkit_path, task_name, task_args):
             task_name, '-Pclargs=' + ','.join(task_args)]
 
 
+def api_name(short_name, version, is_cloud):
```
garrettjonesgoogle: Using a boolean for "is_cloud" is not scalable - it is essentially guaranteed to change as we handle more organizations. I would much prefer an organization identifier, which for this case would be "google_cloud".

pipeline/tasks/gapic_tasks.py
geigerj: Reverted to use the full name.

pipeline/utils/task_utils.py
geigerj: Ah, I don't think I realized there was going to be more of a distinction beyond cloud/non-cloud. Changed this value to "organization_name".

geigerj:
@garrettjonesgoogle I've made the changes requested here and in https://github.com/googleapis/googleapis/pull/215; PTAL.

garrettjonesgoogle:
LGTM

PR #124, opened 2016-12-09
Use "real" googleapis path for MetadataGen

geigerj:
Updates #116 

Previously it was derived from the API name, which is frequently
incorrect. (E.g., "google-cloud-logging-v2" leads to a path of
"google/cloud/logging/v2", even though the real protos are not in
the "cloud" directory".)


garrettjonesgoogle:
LGTM, but might be nice to have @bjwatson check too, because I can't quite tell what the consequences in Toolkit execution are.

geigerj:
@garrettjonesgoogle I'll wait for Brian's approval. At the moment should only change this [line](https://github.com/googleapis/toolkit/blob/6dd1501f1cee45936c3fb600cd172fc4206e69e6/src/test/java/com/google/api/codegen/metadatagen/testdata/python_library.baseline#L263).

It will have a greater impact as we continue to port metadata generation from Packman (which only uses API name) to toolkit (which will have a concept of googleapis_path).

PR #128, opened 2016-12-16
Fix core proto generation

shinfan:
- Core proto pipeline should not require api_version
- Move Java core proto generation from GrpcPipeline into
  CoreProtoPipeline


geigerj:
Does this change mean that the common protos that previously were generated directly by `protoc` now instead use Packman?

shinfan:
@geigerj It should only affect Java common protos

geigerj:
@shinfan Even if it's only for Java, I would prefer not to increase our reliance on Packman unless there's a strong pragmatic reason, since Packman's days are numbered.

Also, could you make sure there's test coverage? It's a little surprising this PR doesn't break the existing tests.

PR #130, opened 2017-01-05
Add package metadata generation task

geigerj:
Note: the TODOs are because this introduces some code duplication/incompatibility with `protoc_tasks`. This duplication/config incompatibility will be resolved in a follow-up PR that switches `GrpcMetadataGen` to use the new package metadata config yaml.

Corresponds to googleapis/toolkit#870

Updates https://github.com/googleapis/toolkit/issues/780


geigerj:
Jumped the gun on this -- need to update tests first.

pipeline/tasks/package_metadata_tasks.py
```diff
@@ -0,0 +1,144 @@
...
+"""Tasks related to package metadata"""
+
+import os
+import yaml
+
+from pipeline.tasks import task_base
+from pipeline.utils import task_utils
+
+
+class PackageMetadataConfigGenTask(task_base.TaskBase):
```
shinfan: I feel like Artman should not directly generate anything but execute through toolkit, shall we move the metadata config generation into the toolkit config gen?

test/testdata/googleapis_test/gapic/packaging/api_defaults.yaml
```diff
@@ -0,0 +1,31 @@
...
+homepage: https://github.com/googleapis/googleapis
+license: Apache-2.0
+semver:
+  # TODO: Python must go to 1.0.20 or 1.1.0 when we GA, due to mistakenly
+  #       publishing some 1.0.x versions to pypi already. They are now
+  #       deleted but files still exist on the pypi servers and can prevent
+  #       us from using the version number again.
+  python:
+    version: '0.14.0'
+    next_version: '0.15.0dev'
```
shinfan: Why not just use `lower_version` and `upper_version`?

geigerj: This will be done in a follow-up (there's a TODO elsewhere in the code to do this). If I do it now, it will break Python gRPC codegen. I want to submit googleapis/toolkit#870 first before I change the gRPC codegen.

pipeline/tasks/package_metadata_tasks.py
geigerj: This is per the discussion in https://github.com/googleapis/toolkit/issues/780; it's not really doing "generation" -- more just merging some existing configs. If the logic ever needs to extend beyond merging a few existing files, I agree that this should be in toolkit.

geigerj: Note that the only transformation happening in this Task is to convert `version`/`next_version` to `lower`/`upper`, which will go away in the follow-up PR.

PR #135, opened 2017-01-27
Don't dump Python classes in package metadata YAML

geigerj:
A user running remote artman encountered a failure where artman dumped
a `!!python/unicode 'v1beta1'` object in the package metadata. This
caused an error when the `toolkit` Java code tried to load the yaml file.

Use `safe_dump` instead of `dump` to ensure that Python objects are
not being dumped in the package metadata.


lukesneeringer:
+1.

PR #137, opened 2017-01-30
Inroduce kubenetes deployment yamls for dev and prod environments. 

ethanbao:
Start the artman conductor as part of the deployment file instead of Dockerfile.

### DIFF
--- a/k8s/dev/dev.yaml
+++ b/k8s/dev/dev.yaml
@@ -1,0 +1,27 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  labels:
+    run: dev
+  name: dev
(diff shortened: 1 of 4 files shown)


k8s/dev/dev.yaml
```diff
@@ -0,0 +1,27 @@
...
+    matchLabels:
+      run: dev
+  template:
+    metadata:
+      creationTimestamp: null
+      labels:
+        run: dev
+    spec:
+      containers:
+      - image: gcr.io/vkit-pipeline/github-googleapis-artman:latest
```
geigerj: Will this be public? Do we want this in our GitHub repo?

PR #139, opened 2017-02-15
Add Python 3 support to Artman

lukesneeringer:
@ethanbao I was going to do some UX hacking on Artman, and discovered that it does not support Python 3. This pull request fixes that.

It does the following:

  * Updates the code to support Python 2.7 and 3.5. This mostly means `print` is consistently used as a function now, and calls to `urllib`, `urlparse`, and `urllib2` now all pull from `six.moves.urllib`, which mirrors the Python 3 package.
  * Substantially refactors the test running. Running tox now does the following, in order:
      * Runs the Python 2.7 tests.
      * Runs the Python 3.5 tests.
      * Runs the linter (on Python 3.5).
      * Puts the coverage data together and produces a report.
  * The `.travis.yml` is also changed to reflect this.


execute_pipeline.py
```diff
@@ -34,20 +34,22 @@
...
 
 """
-
-import sys
+from __future__ import print_function
 import argparse
 import ast
 import base64
 import os
+import sys
```
ethanbao: What is this import for?

lukesneeringer: I just moved it down from a few lines above so it would be in alphabetical order.

ethanbao:
LGTM, thanks for the refactoring!

lukesneeringer:
There is an intermittent failure on one of the tests on Python 3.4 and Python 3.5 only. I am trying to figure out what the issue is.

lukesneeringer:
The intermittent failure is that the test is depending on a dictionary serialization when dictionaries are unordered. Fixing.

PR #142, opened 2017-03-17
Migrate Java core protos generation away from Packman.

shinfan:
Now the Artman is responsible for generating core protos for Java.

```
python execute_pipeline.py \
    --config "../googleapis/gapic/core/artman_core.yaml,\
    ../googleapis/gapic/lang/common.yaml" \
    --language java \
    CoreProtoPipeline
```

TODO(shinfan): Add support in Toolkit to generate package metadata.
TODO(shinfan): Migrate Grpc package tasks


pipeline/tasks/protoc_tasks.py
```diff
@@ -517,3 +517,31 @@ def execute(self, gapic_api_yaml):
...
+            path_components = proto_path.split('/')
+            is_relative_path = False
+            relative_path = []
+            for path_component in path_components:
+                if path_component == 'google':
+                    is_relative_path = True
+                if is_relative_path:
+                    relative_path.append(path_component)
+            output_path = os.path.join(proto_src_dir, '/'.join(relative_path[:-1]))
+            if not os.path.exists(output_path):
```
geigerj: This check is not needed since you are passing the `-p` flag.

pipeline/tasks/protoc_tasks.py
```diff
@@ -517,3 +517,31 @@ def execute(self, gapic_api_yaml):
...
+        for proto_path in src_proto_path:
+            path_components = proto_path.split('/')
+            is_relative_path = False
+            relative_path = []
+            for path_component in path_components:
+                if path_component == 'google':
+                    is_relative_path = True
+                if is_relative_path:
+                    relative_path.append(path_component)
+            output_path = os.path.join(proto_src_dir, '/'.join(relative_path[:-1]))
```
geigerj: `os.path.join(proto_src_dir, *relative_path[:-1])`

(Generally, when referring to actual paths -- i.e., not just parsing the config input -- use only `os.path`, rather than hardcoding the `/` separator.)

pipeline/tasks/protoc_tasks.py
```diff
@@ -517,3 +517,31 @@ def execute(self, gapic_api_yaml):
...
+            if not os.path.exists(output_path):
+                self.exec_command(['mkdir', '-p', output_path])
+            self.exec_command([
+                'cp', '-rf', proto_path, output_path])
+        # Clean up non proto files
+        for root, dirs, files in os.walk(proto_src_dir):
+            for file in files:
```
geigerj: Don't shadow the Python built-in `file`

pipeline/tasks/protoc_tasks.py
```diff
@@ -517,3 +517,31 @@ def execute(self, gapic_api_yaml):
...
+                if not file.endswith('proto'):
```
geigerj: `os.path.splitext(file)[1] == '.proto'`

pipeline/tasks/protoc_tasks.py
```diff
@@ -517,3 +517,31 @@ def execute(self, gapic_api_yaml):
...
                 return go_settings.get('package_name')
+
+class JavaCoreProtoCopyTask(task_base.TaskBase):
+    """Copies the core proto sources to the common-protos output directory
+    """
+    def execute(self, src_proto_path, api_name,
+                api_version, language, organization_name, output_dir):
+        pkg_dir = _prepare_pkg_dir(
+            output_dir, api_name, api_version, organization_name, language)
+        proto_src_dir = os.path.join(pkg_dir, 'protos', 'com')
```
geigerj: This naming is hard to follow: `src_proto_path` is an input, yet `proto_src_dir` seems to be an output?

pipeline/tasks/protoc_tasks.py
shinfan: The file is actually the file name (e.g. Annotation.proto). It feels unnecessary to use os.path in this case.

geigerj: I think `os.path` function is still a nicer in that using a standard method better reveals intent. (And it'll validate `foo.proto` but not `fooproto` or `foo.notproto`.)

shinfan: Ok makes sense.

pipeline/tasks/protoc_tasks.py
shinfan: Renamed to `proto_output_dir`

pipeline/tasks/protoc_tasks.py
shinfan: Done

pipeline/tasks/protoc_tasks.py
shinfan: Done

pipeline/tasks/protoc_tasks.py
shinfan: Done

shinfan:
PTAL

pipeline/tasks/protoc_tasks.py
```diff
@@ -517,3 +517,31 @@ def execute(self, gapic_api_yaml):
...
+    def execute(self, src_proto_path, api_name,
+                api_version, language, organization_name, output_dir):
+        pkg_dir = _prepare_pkg_dir(
+            output_dir, api_name, api_version, organization_name, language)
+        proto_output_dir = os.path.join(pkg_dir, 'protos', 'com')
+        for proto_path in src_proto_path:
+            path_components = proto_path.split('/')
+            is_relative_path = False
+            relative_path = []
+            # Construct the relative proto path e.g. google/api/xxx.proto
```
geigerj: I thought `proto_path` was a directory, not a file? (E.g., https://github.com/googleapis/googleapis/blob/master/gapic/core/artman_core.yaml#L8)

pipeline/tasks/protoc_tasks.py
```diff
@@ -517,3 +517,31 @@ def execute(self, gapic_api_yaml):
...
+            for path_component in path_components:
+                if path_component == 'google':
+                    is_relative_path = True
+                if is_relative_path:
+                    relative_path.append(path_component)
+            output_path = os.path.join(proto_output_dir, *relative_path[:-1])
```
geigerj: Why strip the last element?

pipeline/tasks/protoc_tasks.py
shinfan: You are right. Done.

pipeline/tasks/protoc_tasks.py
geigerj: So if `relative path == ['google', 'api']`, you'd want to `cp -r {src} google` rather than `cp -r {src} google/api`?

shinfan: Ummm I agree that it is a bit confusing but yes in that case the `{src}` directory will be something like `/path/api`
If we use the baseline test as an example, the commands look like
`mkdir -p {CWD}/test/testdata/test_output/google-cloud-library-v1-gen-java/protos/com`
 `cp -rf test/fake-repos/fake-proto {CWD}/test/testdata/test_output/google-cloud-library-v1-gen-java/protos/com`

where the last element is `fake-proto`.
In the `cp` command, the destination should be `.../protos/com` not `.../protos/com/fake-proto`

geigerj: Got it. Just my unfamiliarity with the Java package structure.

pipeline/pipelines/core_generation_pipeline.py
```diff
@@ -71,9 +71,24 @@ def _get_core_codegen_tasks(self, **kwargs):
...
+
+    TODO(shinfan): Add packaging tasks
+    """
+
+    def _get_core_codegen_tasks(self, **kwargs):
+        return [protoc_tasks.ProtoCodeGenTask,
+                protoc_tasks.JavaCoreProtoCopyTask]
+
+    def get_validate_kwargs(self):
+        return ['gapic_api_yaml'] + code_gen.COMMON_REQUIRED
```
geigerj: Why is the `gapic_yaml` needed?

shinfan: I am not sure what this field does and gapic_api_yaml is a field in the artman_core config: https://github.com/googleapis/googleapis/blob/master/gapic/core/artman_core.yaml#L15

I can remove it if it is not necessary.

shinfan: Removed.

PR #145, opened 2017-03-31
Install protobuf with --no-binary

michaelbausor:
Workaround for https://github.com/google/protobuf/issues/2895


Dockerfile
```diff
@@ -57,6 +57,11 @@ ENV SHELL /bin/bash
 RUN yes |ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Linuxbrew/install/master/install)"
 RUN brew doctor || true
 
+# Manually install protobuf with --no-binary flag before installing protoc
+RUN brew install python
```
geigerj: Why is it needed to use a local install of python?

michaelbausor: That is happening anyway when we run `brew install --with-plugins grpc` - we just need to preempt it. I tried to not do this, but then `pip` is stuck at a very early version (which doesn't support the --no-binary arg) and `pip install --upgrade pip` fails (I am not really sure why). Running `pip --version` before and after the `brew install --with-plugins grpc` command shows that it is doing the work of upgrading pip, as part of installing `python`. We move that work to the `brew install python` command.

This could easily be me not understanding docker correctly, and there is something else I should be doing to upgrade pip.

Alternatively, we could try manually uninstalling and reinstalling the protobuf pip package after `brew install --with-plugins grpc`

geigerj: Got it. Not opposed to this, just wondering why, especially if `brew install grpc` is going to lead to switching to linuxbrew python being installed anyway.

Do you mind noting the rationale in a comment? I'm hoping we can stop installing `grpc` with linuxbrew relatively soon, and it might be worth revisiting this then. (Most of the languages already no longer need it, since the codegen in those languages is done through the per-language `grpc-tools` packages.)

PR #150, opened 2017-04-03
Add artman smoketest and integrate it into Dockerfile.

ethanbao:


test/smoketest.py
```diff
@@ -0,0 +1,82 @@
...
+    for f in fnmatch.filter(files, 'artman_*.yaml'):
+      if f in ARTMAN_CONFIG_BLACKLIST:
+        # Do not run generation tests for those in the blacklist.
+        continue
+      filename = os.path.join(root, f)
+      content = open(filename).read()
+      for lang in SUPPORTED_LANGS:
+        if '%s:' % lang in content:
+          if generate_grpc_library(filename, lang):
+            sys.exit(
```
geigerj: Ideally we'd finish running all of the pipelines and report all of their failures before exiting.

test/smoketest.py
```diff
@@ -0,0 +1,82 @@
...
+  'artman_bigtable_admin.yaml'
+]
+
+# Not ready to test Go and Csharp yet.
+SUPPORTED_LANGS = ['python', 'java', 'ruby', 'nodejs', 'php']
+
+def run_smoke_test():
+  artman_config_dir = '../googleapis/gapic/api'
+  for root, dirs, files in os.walk(artman_config_dir):
+    for f in fnmatch.filter(files, 'artman_*.yaml'):
```
geigerj: Note: This assumes that every non-blacklisted artman config will support each of the five supported languages. I *think* that's OK. But in the past we have merged some artman configs into googleapis that only supported a language or two initially, and let the individual languages introduce their configuration as needed. If that happens, it'll cause extraneous test failures until we blacklist the new config.

Dockerfile
```diff
@@ -119,3 +119,7 @@ RUN chmod a+x /usr/local/bin/php-cs-fixer
 ADD . /src
 WORKDIR /src
 RUN pip install -r requirements.txt
+
+# Run smoketests
+# TODO(ethanbao): this should be part of artman CI.
+RUN python test/smoketest.py
```
geigerj: How will the result of this run be reported to trigger/suppress publishing the image to the registry?

ethanbao: If the test failed, it won't be published to container registry. We don't want to publish any docker image that will be broken for sure.

test/smoketest.py
geigerj: Ah, I see this is covered by the `content` check below.

test/smoketest.py
ethanbao: Initially did this as I don't want to wait until everything is done. But it makes sense to capture everything in one round.

ethanbao:
@lukesneeringer

PR #153, opened 2017-04-06
Improve output.

lukesneeringer:
This commit replaces almost all uses of `print` with the `logging` module, and more cleanly separates artman output from the output spit out by the processes that artman runs.

Additionally, it adds a `-v` and `-q` flag to alter how much output you get. Adding `-v` shows all shell commands.


lukesneeringer:
@ethanbao This should make it a bit easier to work on artman; it is the first step toward my GitHub changes. :-) I wanted it to be easier to see what artman is doing.

ethanbao:
This change looks great. One question, artman can also listen to a task queue, fetch task(s) and run tasks in parallel. In addition to the default logger, would it be possible to attach additional stackdriver logger handler for each execution thread, so that each thread will upload the log to cloud with unique log id?

lukesneeringer:
> One question, artman can also listen to a task queue, fetch task(s) and run tasks in parallel.

I have some ideas to how to do this. Will discuss with you later.

lukesneeringer:
@ethanbao has approved in-person. Merging.

lukesneeringer:
I derped and did not see the CI failure. My own fault. Will fix first thing tomorrow morning.

ethanbao:
Oh, I wasn't aware that github will allow merging with CI failure.

lukesneeringer:
It is configurable. We should turn it off. I intend to do that soon on every repo we own.

lukesneeringer:
#154 is the fix. Dumb on me. Feel free to just merge as soon as CI passes.

PR #156, opened 2017-04-12
Add batch generation for grpc pipeline

michaelbausor:
This adds a grpc batch pipeline similar to the gapic batch pipeline. It also adds support for the --stage_output flag for the grpc pipeline, and fixes a bug in batch pipeline generation caused by the introduction of multiple versions of the same API.


garrettjonesgoogle:
The code LGTM - do we need tests?

garrettjonesgoogle:
LGTM

lukesneeringer:
@michaelbausor Could you state how you are using this? I am making some other changes that might affect this and I want to avoid breaking your use case. :-)

michaelbausor:
@lukesneeringer I use the batch mode in a local artman checkout to generate the grpc libraries for all APIs. This is particularly useful in Java because all APIs are kept at the same version, so it is necessary to rerun generation and publish for all of them when the version is bumped.

Additionally, the stage_output flag is also supported. Depending on further discussion, I was planning to add the generated grpc packages to api-client-staging

lukesneeringer:
> Additionally, the stage_output flag is also supported. Depending on further discussion, I was planning to add the generated grpc packages to api-client-staging

I have already implemented this, and it will be in master soon-ish. :-) Letting you know so you do not re-do the work. (You can preview it on `lukesneeringer/github-pr` now if you like.)

michaelbausor:
Cool! Let me know if you want any input re the batch mode, happy to help out.

PR #158, opened 2017-04-14
Update CONTRIBUTING.rst

andreamlin:
Changed one command in instructions.

The current instruction ". ./tox/develop/bin/activate" does not work because 'tox' is not a directory. Replacing it with a command that references a valid path.


garrettjonesgoogle:
Maybe make the commit description more specific.

PR #161, opened 2017-04-20
Support separate proto and grpc generation for Java

michaelbausor:
This PR:
- Separates Java grpc/proto generation into two separate pipelines, ProtoClientPipeline and GrpcClientPipeline. This is required because these are to be two separate packages
- Executes protoc in artman instead of in the generated gradle package


michaelbausor:
Corresponding toolkit PR: https://github.com/googleapis/toolkit/pull/1201

lukesneeringer:
Bias: Python already has this problem, and Artman already solves for it. I would like to know if it would be possible to make Java work more like that.

A couple questions so I can make sure to understand your workflow:

  * Right now (unlike Python), Java does not publish the proto or GRPC packages to api-client-staging. Should that change? (e.g. so we can test it in CI)
  * This means there will now be _three_ distinct packages? (proto, grpc, gapic)
  * Does Java have a local dependency understanding so that the three packages can be tested in tandem in CI?

michaelbausor:
@lukesneeringer 
- Re publishing to api-client-staging: yes, I think we should. But I think that can be in a future PR, as it is not required immediately
- Re 3 packages: yes, that is correct.
- Re local dependency understanding: I'm not fully sure what you're getting at here. We currently don't have any testing around the grpc package, other than a compile check (does any language test the generated proto classes in any way?) Even though we would like to add these packages to api-client-staging, I don't think the intention will be to combine the generation pipelines so that gapic, grpc and proto and all created at the same time. Having the generation of proto and grpc at the same time does make sense though.

michaelbausor:
@lukesneeringer When you say python already has this problem, do you mean the generation of the proto classes and the grpc stubs at the same time? It would be nice for Java to generate both packages in a single pipeline, but we do not want to generate the proto classes and grpc stubs into the same location, and then have to separate them into different packages. Does python handle this sort of separation?

lukesneeringer:
> @lukesneeringer When you say python already has this problem, do you mean the generation of the proto classes and the grpc stubs at the same time? It would be nice for Java to generate both packages in a single pipeline, but we do not want to generate the proto classes and grpc stubs into the same location, and then have to separate them into different packages. Does python handle this sort of separation?

So, for Python the distinction is GAPIC vs. Proto, and yes, it does handle that distinction. You can see a separate `gapic_subpath` and `grpc_subpath` in the Artman config files for Python, and the GAPIC pipeline has code that "rolls in" the GRPC client pipeline for most languages. (I think Java is exempt).

Ideally I would like to follow this pattern for Java, _but_ I also am not going to raise a stink since it is not my workflow.

michaelbausor:
> So, for Python the distinction is GAPIC vs. Proto, and yes, it does handle that distinction. You can see a separate gapic_subpath and grpc_subpath in the Artman config files for Python, and the GAPIC pipeline has code that "rolls in" the GRPC client pipeline for most languages. (I think Java is exempt).
Java is currently exempt, and it seems like it does follow a fairly different workflow here. In Java the grpc packages are updated much less often than the gapic layer (this seems like it would be the case for most languages?). This may change as the gapic layer is stabilized, although I would still expect that changes in gapic yamls will mean that the gapic layer is likely to change significantly more often than the proto/grpc packages.

Having the grpc and proto pipelines combined makes sense. But this runs into some problems using the existing taskflow setup, because the two pipelines share tasks, and some of those tasks "consume" things, like the grpc_code_dir setting. This is not to say that these problems can't be solved, but just running the tasks for these two pipelines sequentially won't work. I think that a fairly natural solution would be to allow the user to specify multiple pipelines to be run from the command line - what do you think?

lukesneeringer:
Sold. I have no objections, then. :-)

lukesneeringer:
Explicit note: Giving you a pass on the lack of tests. :-) I will go back and add them later. (I want to do a full court press to get complete Artman testing, and will do so very soon.)

michaelbausor:
@lukesneeringer OK thanks! I will be happy to help out with test writing

PR #162, opened 2017-04-20
Support multiple git repositories.

lukesneeringer:
This PR addresses a request from @garrettjonesgoogle, who wants to be able to stage both to staging and to the `google-cloud-{lang}` repositories.

This implements that functionality:

  * The `artman_*.yaml` files move from having `git_repo` to `git_repos`, a dictionary which can define one or more. By convention, `staging` must always be defined and is the default.
  * A new `--target-repo` CLI argument is added to specify which repository to stage to.

A connected PR on googleapis will be coming shortly, followed by additional unit tests on this PR.


test/tasks/publish/test_local.py
```diff
@@ -29,7 +29,10 @@ def test_execute(self, exec_command):
         task = local.LocalStagingTask()
         task.execute(
             gapic_code_dir=os.path.expanduser('~/foo/bar'),
-            git_repo={'gapic_subpath': 'pubsub'},
+            git_repo={
+                'gapic_subpath': 'pubsub',
+                'location': 'api_client_staging.git',
```
michaelbausor: I don't think I fully understand the significance of underscores vs. hyphens in this situation. The git repo name is api-client-staging, so why do we need to support conversion from underscores?

artman/cli/main.py
```diff
@@ -290,6 +296,22 @@ def normalize_flags(flags, user_config):
...
+    # Coerce `git_repos` and `target_repo` into a single git_repo.
+    if pipeline_args['publish'] in ('github', 'local'):
+        # Temporarily give our users a nice error if they have an older
+        # googleapis checkout.
+        # DEPRECATED: 2017-04-20
+        # REMOVE: 2017-05-20
+        if 'git_repo' in pipeline_args:
+            logger.error('Your git repos are configured in your artman YAML '
+                         'using a older format. Please git pull.')
+            sys.exit(96)
```
garrettjonesgoogle: How are you choosing exit codes?

test/cli/data/gapic/api/artman_pubsub.yaml
```diff
@@ -17,37 +17,56 @@ common:
...
-  git_repo:
-    gapic_subpath: generated/go/google-pubsub-v1
 csharp:
   gapic_code_dir: ${REPOROOT}/artman/output/gcloud-csharp-pubsub
+go:
+  gapic_code_dir: ${REPOROOT}/gapi-pubsub-go
+  git_repos:
+    go:
+      gapic_subpath: pubsub/apiv1
+      location: git@github.com:GoogleCloudPlatform/google-cloud-go.git
```
garrettjonesgoogle: It would feel more natural for `location` to be before `gapic_subpath`.

Also, `location` is quite generic - maybe `remote_url` would be a better fit.

test/cli/data/gapic/api/artman_pubsub.yaml
```diff
@@ -17,37 +17,56 @@ common:
...
+    staging:
+      gapic_subpath: google-pubsub-v1/cloud.google.com/go/pubsub/apiv1
+java:
+  gapic_code_dir: ${REPOROOT}/artman/output/java/google-cloud-pubsub
+  git_repos:
+    java:
```
garrettjonesgoogle: Maybe instead of naming this after each language, call it `google-cloud` everywhere.

test/tasks/publish/test_local.py
lukesneeringer: I am fairly certain that hyphens are not valid in YAML keys.

artman/cli/main.py
lukesneeringer: This one was arbitrary. (For usage errors I try to stick with the semi-conventional 32.)
I tend to try to make various exits have different codes, especially for tools like this, so things like `subprocess.check_output` can allow some failure codes and not others.

test/cli/data/gapic/api/artman_pubsub.yaml
lukesneeringer: I am indifferent on the ordering (I went with alpha). I do kind of agree that `location` being at the top feels natural, and am happy to do whatever on that.

It would technically be possible to have a non-remote URL.

test/cli/data/gapic/api/artman_pubsub.yaml
lukesneeringer: Honestly, I am happy to let each language lead decide what to name it, and I do not feel they have to match. I plan to send in a googleapis commit so that I am not forcing other people to do the work to accomodate one of my changes,  but I have zero issue if you want to change it.

I am totally happy to name Java's whatever you ask. :-)

For me, personally, I would like them to be named `python` and `nodejs`, because I _do_ work in two languages.

garrettjonesgoogle: For batch work, having the same key would be nice. We already have the ability to distinguish languages using `--language`.

test/cli/data/gapic/api/artman_pubsub.yaml
garrettjonesgoogle: Ahh ok. Maybe `repo_url` then.

test/tasks/publish/test_local.py
lukesneeringer: Just checked: You can have hyphens, but you have to quote the key. I feel that underscores are more readable here.

That said, the spot that you identified is actually an error, and those should be hyphens. (It is not relevant because it is a unit test and the bogus data does not matter; I could have named it foobar.git, but I would like it to match.)

For overriding in `local_paths`, though, it would be underscores. (That said, I expect most people will not override this at all and just let it derive from `reporoot`.)

test/cli/data/gapic/api/artman_pubsub.yaml
lukesneeringer: Are you referring to a case where someone wants to generate a single GAPIC in multiple languages _and stage them to the google-cloud-{lang} repos outright_?

My gut tells me this is not something we want to encourage in the long term. (We want CI from api-client-staging to have that role ultimately. I view the `--target-repo` feature as being primarily necessary not to break the workflow in the meantime.)

That said, there is value in consistency. `google-cloud` feels wrong to me, but I can not come up with something that feels better.

This is easy to change later (it only requires editing the artman YAMLs, not the code), so we have some room to play around on this.

test/cli/data/gapic/api/artman_pubsub.yaml
```diff
@@ -7,47 +7,66 @@ common:
     - google-common-protos
     - google-iam-v1
   import_proto_path:
-    - ${REPOROOT}/googleapis
+    - ${GOOGLEAPIS}
```
ethanbao: Intended to include this change in this PR?

artman/cli/main.py
```diff
@@ -290,6 +296,22 @@ def normalize_flags(flags, user_config):
...
+
+        # Pop the git repos off of the pipeline args and select the
+        # correct one.
+        repos = pipeline_args.pop('git_repos')
+        pipeline_args['git_repo'] = select_git_repo(repos, flags.target_repo)
```
ethanbao: Since target_repo becomes a required flag, we need to make sure it is provided.

test/cli/data/gapic/api/artman_pubsub.yaml
lukesneeringer: Yeah, it was an error / should have been in #159.

artman/cli/main.py
lukesneeringer: It is not a required flag. It defaults to `staging`.

lukesneeringer:
> The feature requirement is not quite clear to me. Is the goal to publish to multiple repos at the same time, or be able to specify target repo? If it is the former, this PR is not implementing that.

It is the latter. (I actively do not want to support the former.)

> For the latter, I want to point out two things: 1) one can override the github repo via CLI flags to publish to a different repo. 2) one can do manual publishing from staging to google-cloud repo, which will always make sure the consistency. If you run the commands in two steps, there are chances that you are generating different files.

1) We can not do that, actually. We only accept username and token flags right now. We could implement this, but the existing implementation has value for the Java team, who has two that they want to be able to switch between easily. Specifying the entire block of information would be cumbersome.

2) This is the ultimate plan, but it is not in place yet, and #159 broke the Java team's workflow in the meantime.

lukesneeringer:
Note: Once I get final signoff on this, I will merge it _and_ release artman 0.3.0.

ethanbao:
Instead of supporting the config at github publish level, would it be better to support at publish step? Basically, normalize the configuration to be publishing config with multiple publishing options (github, package managers), by specify --publish in the command, user specify a concrete publishing item not a publishing type. I feel the new flag is too github specific.

lukesneeringer:
> Instead of supporting the config at github publish level, would it be better to support at publish step? Basically, normalize the configuration to be publishing config with multiple publishing options (github, package managers), by specify --publish in the command, user specify a concrete publishing item not a publishing type. I feel the new flag is too github specific.

I think that is more or less how this is working. Publishers that are repository agnostic ignore this option. (In fact, it does not even get passed in to the final artman args at all.)

lukesneeringer:
I discussed this in person with @ethanbao because I was concerned that I was not understanding his concerns. Summary:

  * Tree-level: He is concerned that other future publishers may benefit from a targetting concept. This is a minor naming change; `--target-repo` shall be renamed `--target`. I agree and will make this change momentarily.
  * Forest-level: One idea he has in mind is to endeavor to make the configuration more linear, and is concerned that the matrix of publisher and target introduced here might make that more challenging. After discussing, we agreed that:
      * There is potential value in configuration simplification, especially when we go third-party.
      * This feature could potentially challenge that, but flattening it prematurely might be over-engineering. We may take a distinct approach when we get to the third-party point.
      * This feature probably only needs to exist in the medium term; our goal is to move publishing to `google-cloud-{lang}` to CI, and then the Java folks should not need to target that repository directory (although maybe they will still have a reason to do so; we can cross that bridge when we get there). Introducing it as-is will allow us to reason about how this particular matrix is used, especially as our reliance on GAPIC-only increases.
      * If, by the time our CI process is in place, there are no other established needs for this, we will probably remove this feature.

Both of us are happy with this plan, so I am changing the spelling of the option and then merging and cutting a release.

PR #164, opened 2017-04-21
Clarify installation steps and usage prerequisites

vchudnov-g:


README.rst
```diff
@@ -41,15 +57,6 @@ For building a GAPIC (the most common task), the usage looks like:
 
     artman --api pubsub --language python
 
-This assumes that you have checkouts of both `googleapis`_ and `toolkit`_
-on your system (and that toolkit is able to run; e.g. you need Java).
-
-Some languages may have additional dependencies; refer to the ``Dockerfile``
-in this repository for canonical installation requirements.
```
lukesneeringer: Why are we removing this segment? It seems important.

README.rst
```diff
@@ -18,16 +18,32 @@ from the protobuf source IDL and additional configuration in YAML files.
 
 Installation
 ------------
+1. Optional: set up a virtualenv in your current directory:
+   .. code::
 
-You can now install artman directly from pip:
+      virtualenv env
```
lukesneeringer: Can we change these instructions to teach `virtualenvwrapper`?

To install and make the virtualenv:

    sudo pip install virtualenv virtualenvwrapper
    mkvirtualenv --python=`which python3` artman

To use the virtualenv later:

    workon artman

This has a few advantages:

  1. Having to track where your virtualenv is located on the filesystem (as the existing instructions require) is clunky. `virtualenvwrapper` solves for this.
  2. These instructions opt you into Python 3 (which is a good thing).

vchudnov-g:
I added `virtualenvwrapper` as the recommended method but kept `virtualenv`. I've liked the fact that it _does_ live in a specific directory, so I don't have to remember which environment goes with which work track I'm on, and so I can get rid of the environment if I decide to delete the work track/directory.

I will definitely try `virtualenvwrapper`, though, because some of the other features are appealing as well.

WDYT?

lukesneeringer:
Sold. Thank you!

PR #168, opened 2017-04-24
Add a few additional unit tests.

lukesneeringer:
I had a few spare cycles this weekend so I decided to inch the unit tests forward a little. Nothing groundbreaking, just gets coverage on a few lines and branches that we did not have before. (Hopefully I send in quite a few more of these off and on over the next couple of weeks.)


test/tasks/publish/test_github.py
```diff
@@ -71,6 +71,57 @@ def test_execute(self, uuid4, chdir, exec_command):
...
+            'git push origin pubsub-python-v1-00000000',
+            'rm -rf /path/to',
+            'rm -rf /tmp/00000000',
+        )
+
+        # Now prove that they were.
+        assert exec_command.call_count == len(expected_commands)
+        for cmd, exec_call in zip(expected_commands, exec_command.mock_calls):
+            _, args, _ = exec_call
+            assert ' '.join(args[0]) == cmd
```
ethanbao: The change looks good to me. Only for my education, is args[0] returning an iterable?

lukesneeringer: Yes, but that is not true by definition. The call signature of `self.exec_command` is to send a single _list_ with essentially `command.split(' ')` (slight oversimplification, but you know what I mean):

```python
self.exec_command(['cp', '-rf', '/src/foo/bar', '/dest/foo/bar'])
self.exec_command(['rm', '-rf', '/foo/bar'])
```

The `args` is just a tuple of the arguments sent to the function, and the first argument (`args[0]`) is an iterable because I send an iterable (a list, in this case) to `exec_command`.

PR #171, opened 2017-04-26
Add per {api x lang} package version config

geigerj:
- Introduce "development status" configuration field
- Allow artman config to override api_defaults file for development
  status and package version to permit these values to be configured
  per {api x lang} combination

See also: https://github.com/googleapis/googleapis/pull/321, https://github.com/googleapis/toolkit/pull/1216


test/testdata/googleapis_test/gapic/packaging/api_defaults.yaml
```diff
@@ -19,6 +19,9 @@ author: Google Inc
 email: googleapis-packages@google.com
 homepage: https://github.com/googleapis/googleapis
 license: Apache-2.0
+development_status:
```
garrettjonesgoogle: We already have a `release_level` concept in gapic config - this seems like the same concept.

geigerj: Ah, I see -- I think you're right. @lukesneeringer, please hold off on review of these PRs while I switch them over to use the existing configuration.

lukesneeringer: Bah, I had just accepted the others. :-(

geigerj:
@lukesneeringer I agree with your assessment that we need better defined merge semantics, and one that is consistent between the package metadata configuration files and the `artman_{api}` files. (In fact, I think if you use the `execute_pipeline.py --config {api x lang},{common}` syntax it does the opposite of what you'd expect, and the common overrides the language-specific. I also agree that that's a separate PR. :)

geigerj:
Tracking as https://github.com/googleapis/artman/issues/175 so we don't forget

PR #172, opened 2017-04-26
Fix a TODO

ethanbao:
Without this, each artman conductor can only take one task.


ethanbao:
Make some additional changes to make the test work, as the while loop won't terminate in test. The change actually reduces the number of mocks in cloudtask_test from 4 to 3 as well.

PTAL

PR #174, opened 2017-04-27
Allow empty commits on --publish=github.

lukesneeringer:
Normally, empty commits are a mistake, but rather than fail the command completely (and in not a nice way), I think it is informative to have the workflow work and check it at commit evaluation time.

Also, for languages with distinct commits for GRPC and GAPIC, it is entirely possible that one will be updated and one will not be, and I do not want a failure in that case.


ethanbao:
LGTM

PR #179, opened 2017-04-28
Add support for batch pipelines

michaelbausor:
This PR:
- Updates the existing batch pipelines to work with the new artman changes
- Adds a `--batch` command line argument
- Adds the JavaProtoCopyTask to include .proto files in Java's proto package

@lukesneeringer Please let me know what additional tests you think are appropriate - I was unsure about how the baseline files in test/testdata are being/going to be used.


artman/tasks/protoc_tasks.py
```diff
@@ -519,3 +519,24 @@ def execute(self, gapic_api_yaml):
...
+    def execute(self, src_proto_path, grpc_code_dir):
+        grpc_proto_dir = os.path.join(grpc_code_dir, 'src', 'main', 'proto')
+        for proto_path in src_proto_path:
+            index = proto_path.rfind('google')
+            if index == -1:
+                raise ValueError('src_proto_path did not contain "google" '
+                                 'in path as expected. src_proto_path: '
+                                 '"{}"'.format(proto_path))
+
+            src_proto_files = self.exec_command(
```
geigerj: Parsing stdout from `find` is a little strange -- could use [`os.walk`](https://docs.python.org/2/library/os.html#os.walk) instead?

(Note: I'm not actually sure where we've set the boundary between using shell commands and using the Python standard library, but ISTM we should prefer the latter.)

artman/tasks/protoc_tasks.py
```diff
@@ -519,3 +519,24 @@ def execute(self, gapic_api_yaml):
                 continue
             if 'package_name' in go_settings:
                 return go_settings.get('package_name')
+
+
+class JavaProtoCopyTask(task_base.TaskBase):
```
geigerj: Did this task disappear in @lukesneeringer's refactoring? I thought I remembered a task by this name existing in the past.

michaelbausor: I was responsible for removing this as part of updating Java proto/grpc generation to run protoc in artman instead of in the generated gradle package. However, this had the side effect of no longer including the raw .proto files as part of the package, so this needs to be added back in.

artman/tasks/protoc_tasks.py
michaelbausor: Agreed, using the standard library is better. Updated.

artman/cli/main.py
```diff
@@ -37,6 +37,7 @@
 from artman.cli.support import parse_github_credentials
 from artman.cli.support import parse_local_paths
 from artman.cli.support import resolve
+from artman.cli.support import select_git_repo
```
michaelbausor: I have updated the imports from `artman.cli.support`. I haven't changed the importing of the `artman.utils.logger` module, as I think any changes there should also update the module itself given that it is designed to be imported in this way.

artman/cli/main.py
```diff
@@ -113,6 +111,14 @@ def parse_args(*args):
              '/path/to/config.yaml:config_section_A|config_section_B. '
              'This is an advanced feature; for most cases, use `--api`.',
     )
+    api.add_argument('--batch',
```
lukesneeringer: How does this interact with publishers? Are the publishing steps done separately or is there an attempt to amalgamate them all into a single publish? (I believe they should remain distinct steps.)

artman/cli/main.py
```diff
@@ -277,6 +275,40 @@ def normalize_flags(flags, user_config):
...
+        flags.config = '{googleapis}/gapic/batch/common.yaml'.format(
+            googleapis=googleapis,
+        )
+        default_pipeline = 'GapicClientBatchPipeline'
+        if not flags.publish:
+            # If publish flag was not set by the user, set it here.
+            # This prevents the user config yaml from causing a
+            # publish event when batch mode is used.
+            flags.publish = 'noop'
+        if flags.target:
```
lukesneeringer: The repo is still taken from the config YAML file when `--target` is used; it just changes _which_ repo. (That said, I agree that batch mode should never be used to publish anywhere other than api-client-staging and api-client-staging-private.)

artman/cli/main.py
```diff
@@ -277,6 +275,40 @@ def normalize_flags(flags, user_config):
...
+                            'the artman config yaml file.')
+            sys.exit(64)
+
+    # Set the pipeline if none was specified
+    if not flags.pipeline_name:
+        flags.pipeline_name = default_pipeline
+
+    # Determine where to publish.
+    pipeline_args['publish'] = support.resolve('publish', user_config, flags,
+        default='github',
```
lukesneeringer: Go ahead and change this to `default='local'`. I got it wrong.

artman/cli/support.py
```diff
@@ -116,3 +116,33 @@ def resolve(name, user_config, flags, default=None):
     if getattr(flags, name, None):
         answer = getattr(flags, name)
     return answer
+
+
+def select_git_repo(git_repos, target_repo):
```
lukesneeringer: I like this move, but make sure the unit tests move also. (The tests directory structure should mirror the code structure. That is not done perfectly right now, but it is done correctly in that portion, and I would like to maintain that.)

test/cli/data/gapic/batch/common.yaml
```diff
@@ -0,0 +1,19 @@
+common:
+  batch_apis: '*'
+  api_config_pattern: ${GOOGLEAPIS}/gapic/api/artman_${API_SHORT_NAME}.yaml
+  artman_language_yaml: ${GOOGLEAPIS}/gapic/lang/common.yaml
+  git_repos:
```
lukesneeringer: I think I would prefer not have an empty dictionary set here. I _want_ the dictionary to fail hard with a not found if you specify one that does not exist in that permutation.

artman/cli/main.py
```diff
@@ -113,6 +111,14 @@ def parse_args(*args):
              '/path/to/config.yaml:config_section_A|config_section_B. '
              'This is an advanced feature; for most cases, use `--api`.',
     )
+    api.add_argument('--batch',
+        action='store_true',
+        help='Enables batch mode, which will generate all APIs for the '
```
shinfan: Before we had the batch support to generate all languages for one API. Is this functionality still available? I feel it is somewhat useful to refresh the API for all languages (e.g. proto update) with one single command.

artman/cli/main.py
shinfan: +1 I agree that the publish should remain a separate step.

shinfan: Instead of having a separate batch mode, why not just use a list in the --api argument?

lukesneeringer: That would be tricky to implement, because `--api` is an alias for a long `--config` argument.
Honestly, if you wanted this, it would be easier to just use a bash loop:

```bash
for API in ( pubsub bigquery vision ) ; do
  artman --api=$API --language=java --publish=github
done
```

artman/cli/main.py
lukesneeringer: Did we really have this functionality? I find that surprising.

This is also something that it might be easier to just use a shell loop:

```bash
for LANGUAGE in ( python java nodejs ruby php go ); do
  artman --api=pubsub --language=$LANGUAGE --publish=github
done
```

artman/cli/main.py
lukesneeringer: Having a batch mode that gives you _all_ APIs has a different value because it opts in new APIs by default.

artman/cli/main.py
michaelbausor: We did have this. It worked in a similar way to how it works with APIs now - the batch pipeline took tasks from each language pipeline.

I have not tried to support it now - I do not think it adds as much value as the API-based batch. Even if we had it, my feeling is that language owners would still want to be responsible for refreshing their own languages, at least in the short term. In the long term, this could all (or nearly all) be automated away by tools that don't rely on a language batch mode.

artman/cli/main.py
michaelbausor: I thought about that approach, but I think having it as a separate flag is clearer, because other CLI behaviour is changed (e.g. changed default pipeline, not accepting the `--target` arg), and having that behaviour change just based on the format of the args to `--api` seems less clear.

Luke's point about opt-in is also a good one.

shinfan: Ok that makes sense

artman/cli/support.py
lukesneeringer: Okay. ^^

artman/pipelines/batch_generation.py
```diff
@@ -103,7 +103,7 @@ def _get_artman_config_filenames(api_config_pattern, batch_apis):
                                                {'API_SHORT_NAME': '*'})
         return sorted(glob.glob(glob_pattern))
     else:
-        if isinstance(batch_apis, basestring):
+        if isinstance(batch_apis, str):
```
lukesneeringer:
```python
if isinstance(batch_apis, (six.text_type, six.binary_type)):
```

michaelbausor: Fixed

PR #181, opened 2017-05-02
Multi paths

lukesneeringer:
This PR generalizes the format in which we accept new paths. It replaces the confusing `gapic_component`, `gapic_subpath`, and `grpc_subpath` with a single list: `paths`. Each item in the list has an artifact (default: `gapic`), src (default: `.`), and dest (default: `.`), with the src and dest being relative to the code generation output location (src) and repository location (dest).

This is a backwards-incompatible change that will require another update to the artman_* files in `googleapis`, which I will submit as a PR shortly. This also means it will be a semver-major bump in Artman, to 0.4.0.


artman/tasks/publish/github.py
```diff
@@ -93,37 +96,33 @@ def execute(self, git_repo, api_name, api_version, language,
 
             # Copy the previously-generated GAPIC into the temporary
             # repository.
-            gapic_subpath = git_repo.get('gapic_subpath', '.')
-            src_path = os.path.abspath(os.path.join(gapic_code_dir, component))
-            self.exec_command(['git', 'rm', '-r', '--force',
-                               '--ignore-unmatch', gapic_subpath])
-            self.exec_command(['cp', '-rf', src_path, gapic_subpath])
-            self.exec_command(['git', 'add', gapic_subpath])
+            for path in git_repo.get('paths', ['.']):
```
ethanbao: Turn this into a util class so that I can be reused in github.py and local.py?

test/cli/data/gapic/api/artman_pubsub.yaml
```diff
@@ -28,45 +28,56 @@ go:
...
       location: git@github.com:GoogleCloudPlatform/google-cloud-go.git
-      gapic_subpath: pubsub/apiv1
+      paths:
+        - dest: pubsub/apiv1
     staging:
-      gapic_subpath: google-pubsub-v1/cloud.google.com/go/pubsub/apiv1
+      paths:
+        - dest: google-pubsub-v1/cloud.google.com/go/pubsub/apiv1
 java:
   gapic_code_dir: ${REPOROOT}/artman/output/java/google-cloud-pubsub
```
ethanbao: Is this still needed?

artman/tasks/publish/github.py
```diff
@@ -55,13 +57,14 @@ def execute(self, git_repo, api_name, api_version, language,
         # ultimately mutate the working directory.
         original_directory = os.curdir
 
-        # Make the GAPIC code directory an absolute path, since we will be
-        # moving around.
-        gapic_code_dir = os.path.realpath(gapic_code_dir)
+        # Track our code directories, and use absolute paths, since we will
+        # be moving around.
+        code_dirs = {'gapic': os.path.abspath(gapic_code_dir)}
```
ethanbao: You plan to replace gapic_code_dir and grpc_code_dir with paths in a follow-up PR?

test/cli/data/gapic/api/artman_pubsub.yaml
lukesneeringer: Yes.

artman/tasks/publish/github.py
lukesneeringer: Yes, I would like to do that.

artman/tasks/publish/github.py
lukesneeringer: If we remove the ability to use a plain string (which I will likely do in a followup), there probably is not need for a util class.

ethanbao: I see

PR #184, opened 2017-05-04
Improve artman remote execution

ethanbao:
Task will be deleted after 3 attempts.
Upload task execution log to cloud logging.
Avoid duplicate stream logging via handlers reset
Use https protocol for github publishing


Dockerfile
```diff
@@ -121,6 +121,10 @@ ADD . /src
 WORKDIR /src
 RUN pip install -e .
 
+# Setup git config used by github commit pushing.
```
lukesneeringer: It is not a big deal here, but it would be wise to get in the habit of thinking a lot about what constitutes a checkpoint in Docker. These commands are making an additional image checkpoint _each_.

I would have expected this to be written:

```Dockerfile
RUN git config --global user.email googleapis-publisher@google.com \
  && git config --global user.name "Googleapis Publisher"
```

This makes one checkpoint instead of two.

artman/conductors/cloudtasks_conductor.py
```diff
@@ -40,21 +46,29 @@ def _pull_and_execute_tasks(task_client, queue_name):
...
+                logger.info('Delete task which exceeds max attempts.')
+                _delete_task(task_client, task)
+                continue
             _execute_task(artman_user_config, task)
             _ack_task(task_client, task)
-            logger.info('Task execution finished for %s' % task)
-        except:
-            logger.error('Task execution failed for %s' % task)
+            logger.info('Task execution finished')
+        except Exception as e:
```
lukesneeringer: This will print the result of `str(e)`, which is just the exception message.
That _might_ be fine in this case (perhaps the message is the output of the command that was run?), but I would probably like to be able to have the full traceback also.

See the [traceback](https://docs.python.org/3/library/traceback.html) module for how to do this; most likely you want something like:

```python
logger.error('\n'.join(traceback.format_tb(sys.last_traceback)))
```

artman/conductors/cloudtasks_conductor.py
```diff
@@ -114,14 +153,16 @@ def _prepare_dir(source_repo="https://github.com/googleapis/googleapis.git"):
     TODO(ethanbao): support loading more input files from heterogeneous data
     sources"""
 
-    repo_root = '/tmp/artman/%s' % str(uuid.uuid4())
+    task_id = str(uuid.uuid4())
```
lukesneeringer: Thought: Would it be better to use `str(uuid.uuid4())[0:8]`? The chances of a collision are still quite low, and it is substantially easier on humans. Considering it is common in git to use only the first seven characters, even with repositories with many, many thousands of commits, I think it would be safe here.

artman/conductors/cloudtasks_conductor.py
```diff
@@ -114,14 +153,16 @@ def _prepare_dir(source_repo="https://github.com/googleapis/googleapis.git"):
...
+    repo_root = '/tmp/artman/%s' % task_id
     logger.info('Prepare a temporary root repo: %s' % repo_root)
     try:
         os.makedirs(repo_root)
     except OSError as e:
         raise e
     logger.info('Checking out fresh clone of %s.' % source_repo)
     googleapis_dir = os.path.join(repo_root, "googleapis")
+    subprocess.check_output(['rm', '-f', '.git/config'])
```
lukesneeringer: TODO: We should move `TaskBase.exec_command` into a general utility function and use it instead of `subprocess.check_output`.

artman/tasks/publish/github.py
```diff
@@ -55,15 +58,18 @@ def execute(self, git_repo, api_name, api_version, language,
         # ultimately mutate the working directory.
         original_directory = os.curdir
 
-        # Make the GAPIC code directory an absolute path, since we will be
-        # moving around.
-        gapic_code_dir = os.path.realpath(gapic_code_dir)
+        # Track our code directories, and use absolute paths, since we will
+        # be moving around.
+        code_dirs = {'gapic': os.path.abspath(gapic_code_dir)}
```
lukesneeringer: This stuff looks like changes I made. You probably need to merge master into this.

artman/conductors/cloudtasks_conductor.py
ethanbao: Done.

artman/conductors/cloudtasks_conductor.py
ethanbao: https://github.com/googleapis/artman/issues/185

artman/conductors/cloudtasks_conductor.py
lukesneeringer: Were you able to update this? (This is the only other concern I have on this PR.)

lukesneeringer: Nevermind, slow plane Internet means I did not see your comment. Approving!

PR #187, opened 2017-05-06
Remove default path

michaelbausor:
I don't know if this is the right fix, but I wanted to submit a PR rather than just file an issue. Currently if no path is specified, using a default of `'.'` will cause the generated package to replace the entire local `api-client-staging` repo (when staging is the publish location). This could potentially cause loss of work, and can happen when running generation for any API with paths set - currently including DLP and IAM.

This also updates the configs with the latest changes.


artman/tasks/publish/local.py
```diff
@@ -73,7 +73,7 @@ def execute(self, gapic_code_dir, git_repo, local_paths,
         dests = []
 
         # Determine where the code belongs and stage it there.
-        for path in git_repo.get('paths', ['.']):
```
lukesneeringer: Empty list is definitely not what we want, whereas `['.']` does make sense.
I understand what you are saying about the "opt-in" problem, though.

Maybe we should make it an error if `paths` is not set? (This proposal amounts to failing silently.)

michaelbausor: I think that an error is a good idea. We should generate an error here, but probably also we could check for this error during pipeline construction, WDYT?

lukesneeringer: I am not particularly worried about catching the error as soon as possible in the chain, but I am not opposed to it if it is easy. Definitely change this to a flat error though:

```python
if 'paths' not in git_repo:
    raise RuntimeError('This git repository entry in the artman YAML does not define module paths.')
```

Note that if you do this, you need to go back and add a `.` path to the spots where it is appropriate.

michaelbausor: Changed to raise an error. I am not sure what spots need to be changed to add the `.` path - I can't see any usage that relies on it (for cases such as DLP where it is currently missing, we want to fail).

PR #191, opened 2017-05-18
Fix protoc generation task

shinfan:
- Separate metadata generation tasks for gRPC and Proto packages.
- Fix an issue in JavaProtoCopyTask where grpc_code_dir was incorrectly used.


artman/pipelines/grpc_generation.py
```diff
@@ -118,7 +118,7 @@ def get_grpc_codegen_tasks(self, **kwargs):
             protoc_tasks.ProtoDescGenTask,
             protoc_tasks.ProtoAndGrpcCodeGenTask,
             package_metadata_tasks.PackageMetadataConfigGenTask,
-            package_metadata_tasks.PythonProtoPackageMetadataGenTask,
```
lukesneeringer: Please make sure you test this. The original `ProtoPackageMetadataGenTask` change broke my grpc packages (they were copied with no metadata).

artman/tasks/package_metadata_tasks.py
```diff
@@ -94,12 +95,42 @@ def _write_yaml(self, config_dict, dest):
...
+
+        return config
+
+# Metadata gen
+
+class ProtoPackageMetadataGenTaskBase(task_base.TaskBase):
+    def _execute(self, api_name, api_version, organization_name, local_paths,
                 descriptor_set, src_proto_path, service_yaml,
-                grpc_code_dir, output_dir, package_metadata_yaml,
+                input_dir, output_dir, package_metadata_yaml,
```
michaelbausor: Where is the value of input_dir coming from?

artman/tasks/package_metadata_tasks.py
```diff
@@ -94,12 +95,42 @@ def _write_yaml(self, config_dict, dest):
         with open(dest, 'w') as f:
             yaml.safe_dump(config_dict, f, default_flow_style=False)
 
-class ProtoPackageMetadataGenTask(task_base.TaskBase):
-    default_provides = 'proto_pkg_dir'
+class JavaGrpcPackageMetadataConfigGenTask(PackageMetadataConfigGenTask):
```
michaelbausor: Is it necessary to move these here, instead of leaving them in place? (Not a big deal, just a bit harder to see what has changed)

artman/tasks/package_metadata_tasks.py
shinfan: This `ProtoPackageMetadataGenTaskBase` becomes a base class of `GrpcPackageMetadataGenTask` and `ProtoPackageMetadataGenTask`. The input_dir is either `proto_code_dir` or `grpc_code_dir` depending on the caller

artman/tasks/package_metadata_tasks.py
shinfan: I was trying to group the `config gen task` and `metadata gen task` :)
See L24 and L128

artman/pipelines/grpc_generation.py
shinfan: I will do a manual test to make sure it works for Python but I feel this maybe the case that the Python unit tests should catch.

artman/tasks/package_metadata_tasks.py
michaelbausor: Ah OK thanks.

michaelbausor:
I am not completely sure I am reading the PR correctly - is the fix in the JavaProtoCopyTask the only change in output? Or do the changes in package_metadata_tasks also affect output?

shinfan:
@michaelbausor This PR fixes two issues:
1) JavaProtoCopyTask fixes the issue that proto packages contains gRPC clients
2) package_metadata_tasks fixes 'grpc_code_dir' exception (becasue proto metadata should depend on proto_code_dir)

artman/pipelines/grpc_generation.py
shinfan: I have manually verified the grpc pipeline and it output looks good

shinfan:
@lukesneeringer PTAL

artman/pipelines/grpc_generation.py
lukesneeringer: Great. Thank you.

PR #192, opened 2017-05-18
Support local publish for Grpc/ProtoClientPipeline

shinfan:
Currently Artman does not support local publish for Grpc/ProtoClientPipelines. This causes manual work if the user wants to stage the proto packages into api-client-staging.

This PR enables `LocalStagingTask` for `grpc_generation` module. The command looks like:

```
artman --config ../googleapis/gapic/core/artman_core.yaml,../googleapis/gapic/lang/common.yaml --language java --pipeline ProtoClientPipeline --publish=local
``` 

This PR also allows the default `GapicClientPiepline` to produce non-gapic packages such as common-protos and iam. It will only stage the proto/grpc packages if the target API does not have GAPIC client. As a result the following command will work:
```
artman --config ../googleapis/gapic/core/artman_core.yaml,../googleapis/gapic/lang/common.yaml --language java
```
This is a pre-requisite change of enabling batch generation of staging packages. I will send a follow-up PR that enables one-command Java staging after this one is merged.


michaelbausor:
LGTM

PR #193, opened 2017-05-19
Smaller docker image

lukesneeringer:
This new `Dockerfile` should provide a smaller image (~although I am unsure at the moment exactly _how much_ smaller~ 1.3 GB smaller), at the expense of a slightly longer build time.


lukesneeringer:
Tagging this as `do not merge` until my local build finishes and I can get a final size comparison.

lukesneeringer:
Update: The size reduction goes from 3.4 GB to 2.5 GB.

I am fixing two additional things:

  * [ ] Moving from Python 2 to Python 3.
  * [ ] Fixing the Ruby GRPC pipeline.

lukesneeringer:
Okay, I have made several more significant improvements:

  * Use Ubuntu 16.04 (not 14.04)
  * This also means we can install Ruby, Node, and Go via. `apt-get`, because Ubuntu 16.04 has new enough versions. Linuxbrew is gone.
  * Ruby is now on 2.3 (previously it was 1.9).
  * Artman is now installed from pip.

lukesneeringer:
@ethanbao This is now ready for review.

Dockerfile
```diff
@@ -1,13 +1,15 @@
...
+FROM ubuntu:16.04
+
+ENV DEBIAN_FRONTEND noninteractive
 
 # Set the locale
-RUN locale-gen en_US.UTF-8
 ENV LANG en_US.UTF-8
 ENV LANGUAGE en_US:en
-ENV LC_ALL en_US.UTF-8
+ENV LC_ALL C
```
lukesneeringer: This is required for pip to work properly. Also, `C` is the correct value. :-)

Dockerfile
```diff
@@ -18,117 +20,129 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     curl \
     kdiff3 \
     git \
-    vim # for debug
+    vim \
+  && rm -rf /var/lib/apt/lists/*
```
lukesneeringer: You are going to see `  && rm -rf /var/lib/apt/lists/*` a lot (after any `apt-get` call). This is to wipe out its index so it is not saved to the image, which is where we are getting a decent chunk (about 800 MB) of the size savings.

Dockerfile
```diff
@@ -18,117 +20,129 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
-RUN add-apt-repository ppa:openjdk-r/ppa
-RUN apt-get update && apt-get install -y openjdk-8-jdk
+RUN add-apt-repository ppa:openjdk-r/ppa \
+  && apt-get update \
+  && apt-get install -y openjdk-8-jdk \
+  && rm -rf /var/lib/apt/lists/*
 
 # Setup JAVA_HOME, this is useful for docker commandline
 ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64/
 RUN export JAVA_HOME
```
lukesneeringer: I expect this line is superfluous but I never tested.

Dockerfile
```diff
@@ -18,117 +20,129 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
 
-# Install NodeJS
-ADD ./scripts /scripts
-WORKDIR /scripts
-# This is need to install nodejs 4.x otherwise nodejs 0.x will be installed.
-RUN bash setup_node4.sh
-RUN apt-get install -y nodejs
-
-# Install linuxbrew.
-RUN useradd -m -s /bin/bash linuxbrew
```
lukesneeringer: I was able to completely nix Linuxbrew because Ubuntu 16.04 has a new enough version of everything, meaning we no longer incur the overhead of an entirely new package manager. This saved about 200 MB.

Dockerfile
```diff
@@ -18,117 +20,129 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
-
-USER root
+# Install NodeJS.
+# This installs Node 4 on Ubuntu 16.04.
+RUN apt-get update \
+  && apt-get install -y \
+    nodejs \
+    npm \
+  && rm -rf /var/lib/apt/lists/* \
+  && ln -s /usr/bin/nodejs /usr/local/bin/node
```
lukesneeringer: This `ln -s` is because the Node.js package on apt is quirky and uses `nodejs` instead of `node` as the executable.

Dockerfile
```diff
@@ -18,117 +20,129 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
 # TODO: consider installing released packages once artman is versioned (so that
 #   each release of artman is pegged to a release of packman)
 RUN npm install -g https://github.com/googleapis/packman.git
 
 # Setup tools for codegen of Ruby
-RUN gem install rubocop --version '= 0.39.0' --no-ri --no-rdoc
-RUN gem install bundler --version '= 1.12.1' --no-ri --no-rdoc
-RUN gem install rake --version '= 10.5.0' --no-ri --no-rdoc
-RUN gem install grpc-tools --version '=1.0.0' --no-ri --no-rdoc
+RUN gem install rake --no-ri --no-rdoc \
```
lukesneeringer: Installing `rake` was required for the `rubocop` to work with this version of Ruby.

Dockerfile
```diff
@@ -18,117 +20,129 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
-WORKDIR /
 RUN git clone https://github.com/googleapis/googleapis
 RUN git clone https://github.com/googleapis/toolkit
 ENV TOOLKIT_HOME /toolkit
-WORKDIR /toolkit
-RUN sudo ./gradlew install # Install toolkit. Must sudo to download gradle plugins.
+
+# Install toolkit. Must sudo to download gradle plugins.
+RUN cd /toolkit \
+  && ./gradlew install \
```
lukesneeringer: sudo is no longer needed. :-)

Dockerfile
```diff
@@ -18,117 +20,129 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
 ADD artman-user-config-in-docker.yaml /root/.artman/config.yaml
 
-# Run smoketests
-# TODO(ethanbao): this should be part of artman CI.
-RUN python test/smoketest.py
+# Install artman and run the smoke test.
+RUN pip3 install googleapis-artman==0.4.3
+
+# TODO (lukesneeringer): Move smoke tests to a different venue.
+# RUN smoketest_artman.py \
```
lukesneeringer: I commented this out for now; I am not convinced that the `docker build` is the right place to do this.

You can always run it yourself with:

```
docker run -it googleapis/artman:0.4.3
smoketest_artman.py
```

artman/tasks/protoc_tasks.py
```diff
@@ -240,7 +240,7 @@ class RubyGrpcCopyTask(task_base.TaskBase):
     the gapic_code_dir/lib.
     """
     def execute(self, api_name, api_version, language, organization_name,
-                output_dir, gapic_code_dir, grpc_code_dir):
+                output_dir, pkg_dir, gapic_code_dir, grpc_code_dir):
```
lukesneeringer: I added this to try and solve the `NameError` farther down, but the problem is deeper. It does need to be there though.

setup.py
```diff
@@ -27,7 +27,7 @@
 
 setuptools.setup(
     name='googleapis-artman',
-    version='0.4.1',
+    version='0.4.3',
```
lukesneeringer: I published `0.4.2` intermittedly by mistake.

test/smoketest_artman.py
```diff
@@ -60,7 +60,6 @@ def run_smoke_test():
 def generate_grpc_library(artman_config, lang):
     grpc_pipeline_args = [
         'artman',
-        '--user-config', 'test/artman_user_config_in_smoketest.yaml',
```
lukesneeringer: This was incompatible with the Docker image; it was working previously by mistake because the local artman directory was being added to the image.

lukesneeringer:
@ethanbao I put a lot of comments in to preemptively explain why some items have changed, since I figured you would want to know. :-)

test/smoketest_artman.py
ethanbao: The test won't work after removing this, right?

Dockerfile
```diff
@@ -18,117 +20,129 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
     software-properties-common \
-    php-pear
+    php-pear \
+    python3-pip \
+  && rm -rf /var/lib/apt/lists/*
+
+# Install protoc 3.3.0.
+# Currently, the stable version is still 2.6.x, which can only handle proto2
+# syntax, so we have to download our own.
+RUN mkdir -p /usr/src/protoc/ \
```
ethanbao: nit: I feel protos upgrades more often than JDK, consider move protoc installation after language tool installation (JDK, ruby, etc).

Dockerfile
ethanbao: Ack. That's lots of save.

Dockerfile
ethanbao: Can you make this comment as part of the comment in Dockerfile?

Dockerfile
```diff
@@ -18,117 +20,129 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
+  && apt-get update \
+  && apt-get install -y openjdk-8-jdk \
+  && rm -rf /var/lib/apt/lists/*
 
 # Setup JAVA_HOME, this is useful for docker commandline
 ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64/
 RUN export JAVA_HOME
 
-# Install NodeJS
-ADD ./scripts /scripts
```
ethanbao: Great. You can then remove the scripts.

Dockerfile
```diff
@@ -18,117 +20,129 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
+
+# Install Go.
+# This installs Go 1.6 on Ubuntu 16.04.
+RUN apt-get update \
+  && apt-get install -y golang \
+  && rm -rf /var/lib/apt/lists/*
 
+# Download the go protobuf support.
 ENV GOPATH /go
 ENV PATH $GOPATH/bin:/home/linuxbrew/.linuxbrew/opt/go/libexec/bin:$PATH
```
ethanbao: You can remove the linuxbrew part from $PATH.

Dockerfile
ethanbao: Woohoo.

Dockerfile
```diff
@@ -18,117 +20,129 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
-
-# Install artman.
-# TODO(ethanbao): pipeline should be installed from package manager.
-ADD . /src
-WORKDIR /src
-RUN pip install -e .
+RUN pip3 install -e git+https://github.com/googleapis/proto-compiler-plugin#egg=remote
+
+# Install PHP protobuf plugin.
+RUN gem install ronn --no-ri --no-rdoc \
```
ethanbao: Just curious, will gem installation introduces intermediate data like apt-get?

setup.py
```diff
@@ -42,6 +42,9 @@
             'start-artman-conductor = artman.cli.conductor:start',
         ],
     },
+    scripts=[
+        'test/smoketest_artman.py',
```
ethanbao: Just to make sure I follow the release process. This script will be executed every time a new artman pypi package (plus artman docker image) gets released?

Dockerfile
lukesneeringer: Yes.

test/smoketest_artman.py
lukesneeringer: It _will_ work now that this is removed. It would not work with it present.

Dockerfile
lukesneeringer: Done.

Dockerfile
lukesneeringer: gem does not (to the best of my knowledge) maintain a running list of all available packages in the same way that apt does.

When you run `apt-get update`, it literally pulls down metadata for _every single available package_ on those apt repos, and then uses that when you run `apt-cache search` or `apt-get install`.

I suspect the reason for this behavior is that this is where the multiple sources of truth (the N apt repositories you have added to `sources.list`) is resolved into a single list of packages; it would be too cumbersome to check each separate repo for a package every time you ran an install (especially given the high dependency count of apt packages).

By contrast, `gem install` (and pip and npm) check a single source of truth, and so no complex local cache is required.

setup.py
lukesneeringer: Yes, and that is a manual process in the immediate term.

PR #197, opened 2017-05-24
Support excluded_proto_path in proto tasks

shinfan:
Artman config PR: https://github.com/googleapis/googleapis/pull/355

Fixes https://github.com/googleapis/googleapis/issues/353


shinfan:
Friendly ping

artman/utils/protoc_utils.py
```diff
@@ -298,11 +298,25 @@ def prepare_pkg_dir(output_dir, api_name, api_version, organization_name,
     return pkg_dir
 
 
-def find_proto_files(proto_path):
-    for dirpath, subdirs, files in os.walk(proto_path):
-        for f in files:
-            if f.endswith('.proto'):
-                yield os.path.join(dirpath, f)
+def find_protos(proto_paths, excluded_proto_path):
+    """Searches along `proto_path` for .proto files and returns a generator of
```
michaelbausor: Nit: change `proto_path` to `proto_paths`

artman/utils/protoc_utils.py
```diff
@@ -298,11 +298,25 @@ def prepare_pkg_dir(output_dir, api_name, api_version, organization_name,
...
+    paths"""
+    if type(proto_paths) is not list:
```
michaelbausor: An alternative would be to check if proto_paths was a stringy type, and error then (or accept the single item). Then this would accept generators. But, am fine to leave as is

test/fake-repos/fake-proto/excluded/excluded.proto
```diff
@@ -0,0 +1,8 @@
+// This is a fake proto file for testing.
```
michaelbausor: How does the test know whether this proto was excluded or not? Because it contains no content, would the test have the same result if this was included?

artman/utils/protoc_utils.py
lukesneeringer:
```python
if isinstance(proto_paths, (collections.Generator, collections.MutableSequence)):
```

lukesneeringer: You will need to `import collections` too, obviously.

test/fake-repos/fake-proto/excluded/excluded.proto
shinfan: I have added test cases that check for the returned value of find_proto methods so the content of the proto should not matter. Consider done.

artman/utils/protoc_utils.py
shinfan: Python complains that collections module does not have `Generator` so I use `types.GeneratorType` instead. Consider done.

artman/utils/protoc_utils.py
shinfan: Done

shinfan:
@lukesneeringer @michaelbausor Sorry I ended up with a force push since my master branch history was a bit messy. PTAL.

PR #199, opened 2017-05-25
Decode output from check_output

shinfan:


ethanbao:
Is this change needed because artman now uses python3?

shinfan:
@ethanbao Yes, check_output returns a byte string. In Python 2.7 this does not cause problems but Python 3 handles byte string differently from normal string, which led to the "program not found" error.

PR #202, opened 2017-05-26
Dockerfile edits.

lukesneeringer:


Dockerfile
```diff
@@ -103,11 +103,16 @@ RUN gem install rake --no-ri --no-rdoc \
   && gem install grpc-tools --version '=1.0.0' --no-ri --no-rdoc
 
 # Install couple of git repos
-RUN git clone https://github.com/googleapis/googleapis
-RUN git clone https://github.com/googleapis/toolkit
+RUN git clone https://github.com/googleapis/googleapis \
```
ethanbao: nit: move it below the Protobuf-PHP installation.

Dockerfile
```diff
@@ -45,7 +45,7 @@ RUN mkdir -p /usr/src/protoc/ \
 # Install GRPC and Protobuf.
 RUN pip3 install --upgrade pip \
   && pip3 install \
-    grpcio==1.3.0 \
+    grpcio==1.3.5 \
```
ethanbao: I am not sure whether the grpcio version here must match the one in artman requirements.txt file. If that is the case, let's add a comment.

lukesneeringer: It is not the end of the world if it does not; artman will uninstall the old one and install the new one. It makes the image slightly smaller if they match. :-)

PR #204, opened 2017-05-26
Two small fixes

ethanbao:
Include csharp into smoketest.
include cloudtasks.json into the package.


lukesneeringer:
This is gated on merging #203, which just requires approval.

lukesneeringer:
Superceded by #208.

PR #205, opened 2017-05-28
Update batch mode API list

michaelbausor:
Update batch mode API list to handle api_config_patterns


lukesneeringer:
Supreceded by #206.

PR #208, opened 2017-06-02
Smoke Test (non-Fork)

lukesneeringer:
This is a non-fork equivalent to #204.


test/smoketest_artman.py
```diff
@@ -36,7 +36,7 @@
 ]
 
 # Not ready to test Go and Csharp yet.
-SUPPORTED_LANGS = ['python', 'java', 'ruby', 'nodejs', 'php', 'go']
+SUPPORTED_LANGS = ['python', 'java', 'ruby', 'nodejs', 'php', 'go', 'csharp']
```
ethanbao: I think it is okay to list csharp now even though the layout will be changed. You will need to pull the latest code and remove one TODO before you can merge this PR.

@jskeet FYR

PR #210, opened 2017-06-03
Support proto_test_deps in metadata config

shinfan:
This `proto_test_deps` field is currently used by the Java gapic build files to declare test dependencies.


shinfan:
@garrettjonesgoogle  Can you `approve` the change? I don't think Github recognize 'LGTM' as approval :)

PR #211, opened 2017-06-05
cache protobuf_path

pongad:
Batch jobs previously called `./gradlew showProtobufPath` many times.
This PR caches the result.
On my machine, this reduces the run time of `artman --batch --language java --publish noop`
from 5 minutes to 3:15.


artman/utils/protoc_utils.py
```diff
@@ -305,9 +305,12 @@ def find_proto_files(proto_path):
             if f.endswith('.proto'):
                 yield os.path.join(dirpath, f)
 
-
+_protobuf_path = None
```
pongad: Not sure if there's a better way to do this in Python?

PR #213, opened 2017-06-06
Fix a bug related artman/csharp-grpc generation in Dockerfile.

ethanbao:
Download and setup grpc_csharp_plugin in Dockerfile.
Enable csharp/grpc generation in circleci config.


.circleci/config.yml
```diff
@@ -79,8 +79,14 @@ jobs:
           name: Build Pub/Sub (C#)
           command: artman --api=pubsub --language=csharp --publish=noop
       - run:
+          name: Build Pub/Sub (C#/gRPC)
+          command: artman --api=pubsub --language=csharp --pipeline=GrpcClientPipeline --publish=noop
```
lukesneeringer: Can we make C# work like the other languages where the GAPIC pipeline includes the GRPC pipeline? (This just requires removing my two-line `if` block that circumvented this behavior.)

Dockerfile
```diff
@@ -49,6 +49,12 @@ RUN pip3 install --upgrade pip \
     grpcio==1.3.5 \
     protobuf==3.3.0
 
+# Install grpc_csharp_plubin
+RUN curl -L https://www.nuget.org/api/v2/package/Grpc.Tools/1.3.6 -o temp.zip
```
lukesneeringer: This needs to be one `RUN` command:

```Dockerfile
RUN curl -L https://www.nuget.org/api/v2/package/Grpc.Tools/1.3.6 -o temp.zip \
  && unzip -p temp.zip tools/linux_x64/grpc_csharp_plugin > /usr/local/bin/grpc_csharp_plugin \
  && chmod +x /usr/local/bin/grpc_csharp_plugin \
  && rm temp.zip
```

Otherwise the intermediate artifacts remain forever.

lukesneeringer:
@ethanbao I want to block on fixing the `RUN` command, but happy to approve with the GAPIC/GRPC generation for C# as-is.

.circleci/config.yml
ethanbao: Yes, this can be removed after Jon finishes the csharp/artman integration.

ethanbao:
Shin/Luke, please also review the change in python_grpc_tasks.py

.circleci/config.yml
```diff
@@ -81,6 +81,14 @@ jobs:
       - run:
           name: Build Speech (C#)
           command: artman --api=speech_v1 --language=csharp --publish=noop
+      # TODO(jskeet): Remove C#/gRPC generation tests after Cshape gapic
```
jskeet: Nit: Cshape => csharp :)

CONTRIBUTING.rst
```diff
@@ -120,6 +120,29 @@ Running Tests
...
 
+-   To run the full set of artman smoke tests, you need to install docker
+    installed on your machine, and run the following at the root of your
+    artman source directory:
+
+  .. code::bash
+
+      docker pull googleapis/artman:latest
+      docker run -it \
+      -v ${PWD}:/usr/src/artman \
```
lukesneeringer: Not a blocker, but a two-space hanging indent on line-continued commands would be nice.

.circleci/config.yml
```diff
@@ -81,6 +81,14 @@ jobs:
...
+      # generation includes gRPC generation.
```
jskeet: What does that involve? Is there any particular benefit to combining two steps which use completely different generators?

.circleci/config.yml
ethanbao: Done

.circleci/config.yml
ethanbao: That is mainly because gapic wrapper alone won't work without grpc library. For external artman users, I would envision them to use artman to generate one self-contained working library with one single command, without knowing the relation between gapic and grpc.

PR #217, opened 2017-06-08
Add Python documentation

lukesneeringer:
This PR adds [protoc-docs-plugin](https://github.com/googleapis/protoc-docs-plugin) support for the Python proto generation.

(It is also a test of whether CircleCI is working against forks now.)


Dockerfile
```diff
@@ -154,8 +164,4 @@ RUN mkdir -p /root/
 ADD artman-user-config-in-docker.yaml /root/.artman/config.yaml
 
 # Install artman and run the smoke test.
-RUN pip3 install googleapis-artman==0.4.4
-
-# TODO (lukesneeringer): Move smoke tests to a different venue.
-# RUN smoketest_artman.py \
-#   && rm -rf /artman/output
+RUN pip3 install googleapis-artman==0.4.5
```
ethanbao: I will drop my PR which does the same change here.

Dockerfile
```diff
@@ -124,6 +124,16 @@ RUN gem install ronn --no-ri --no-rdoc \
   && chmod a+x /usr/local/bin/php-cs-fixer \
   && cd /
 
+# Set up tools for Python codegen; these are:
+#   pandoc: an apt package that can convert text between formats
+#     (example: Markdown to Restructured Text)
+#   protoc-docs-plugin: A protoc plugin to add docstrings to the Python
+#     protoc output.cd
+RUN apt-get update \
+  && apt-get install -y pandoc \
```
ethanbao: Is "pandoc" used anywhere in artman code? github codesearch doesn't tell me anything.

lukesneeringer: It's used by protoc-docs-plugin. (It converts Markdown to ReST.)

Dockerfile
lukesneeringer: Ha! Sorry. :-p

Dockerfile
ethanbao: Does "--pydocstring_out={root}" flag below imply the use of protoc-docs-plugin?

lukesneeringer: Exactly. The way protoc works is that `--foo_out={place}` translates to running `protoc-gen-foo` in `$PATH`.

PR #221, opened 2017-06-15
update toolkit

pongad:
Update toolkit to latest version to
bring in changes from https://github.com/googleapis/toolkit/pull/1319.


ethanbao:
The current smoketest running in circle CI catch potential issue with this Dockerfile change. Can you run the following in your local machine, and make sure the local toolkit is synced to the exact commit:

docker run -it \
        -v ${PWD}:/usr/src/artman \
        -v {REPLACE_WITH_LOCAL_TOOLKIT_DIR}:/toolkit \
        googleapis/artman:latest \
        /bin/bash -c  "pip uninstall -y googleapis-artman; \
        pip install -e /usr/src/artman/; \
        /usr/src/artman/test/smoketest_artman.py --apis=pubsub,vision"

pongad:
@ethanbao I think this is because an update to toolkit requires a change in `googleapis`. If I also add a 
```shell
-v $GOOGLEAPIS_PATH:/googleapis
```
everything works OK. Does this complicate matters?

For my own information, how much work is it to sync after the dockerfile is changed? If the work is significant, I'll try to batch changes in the future.

ethanbao:
Your change is good. Use of commit hash is a temporary workaround because toolkit itself doesn't have its own release yet. The problem will eventually be how toolkit decides the cut of its release.

PR #222, opened 2017-06-19
Bump the artman version to 0.4.7

ethanbao:


ethanbao:
Reviewers, the smoke-java test is known to fail and will pass after googleapis/artman:0.4.7 is released to docker hub.

PR #227, opened 2017-06-20
Update the Artman pipelines for C#

jskeet:


Dockerfile
```diff
@@ -140,7 +140,7 @@ RUN git clone https://github.com/googleapis/googleapis \
   && rm -rf /googleapis/.git/
 RUN git clone https://github.com/googleapis/toolkit \
   && cd toolkit/ \
-  && git checkout 490ed7f433d12649fd508b2d4c84fb65c3357955 \
+  && git checkout 7a597cdef2d6da40349cf820498226c286ef76c8 \
```
jskeet: Yes, it's merged - it's a few commits after my one, basically.

jskeet:
Thanks - I ran the smoke test manually and it was fine. If you'd like me to send one PR for the commit update, and then do the others separately, that would be fine.

Dockerfile
ethanbao: My bad. The snippet in the github commit history is the leading hash not the ending one.

PR #232, opened 2017-06-29
Set include_package_data to true so that non-code-file can also be installed

ethanbao:
Bump the pypi version to 0.4.9


artman/conductors/cloudtasks_conductor.py
```diff
@@ -75,8 +75,8 @@ def _pull_and_execute_tasks(task_client, queue_name):
 
 def _create_tasks_client():
     credentials = GoogleCredentials.get_application_default()
-    with open(
-        os.path.join(os.path.dirname(__file__), 'cloudtasks.json'), 'r') as f:
+    with open(os.path.join(os.path.dirname(__file__), 'cloudtasks_json.py'),
```
lukesneeringer: This is not quite what I meant. Now you are just loading a Python file (that also happens to be valid JSON) and loading it that way. Instead, just _import_ the Python file. (If you really need JSON for some reason, call `json.dumps` on the object you import.)

The reason for this (besides getting around your MANIFEST.in issue, which could be solved other ways) is that importing is faster/smarter/better than a straight-up load from disk.

PR #235, opened 2017-07-10
Bump artman version to 0.4.10

ethanbao:


michaelbausor:
What versioning strategy are we using for artman? The most recent change I introduced does not change the CLI, but it is potentially "breaking" because it removes the requirement for packman for PHP generation, but adds a requirement for the PHP grpc protoc plugin. Given that, do we need to bump to 0.5.0?

ethanbao:
We haven't set any restricted rule, given the fact that we don't have external customers yet. So far, we bump the minor version when there are major changes to artman code itself (not the language-specific code).

ethanbao:
I just re-enabled the php smoketest in circleci, since the test should pass now.

PR #237, opened 2017-07-13
Include .NET Core SDK in Dockerfile

jskeet:


ethanbao:
Can you run `docker images` and check the size difference between the one with this change vs image version 0.4.10?

jskeet:
Will do in the morning. I expect it'll be a pretty large size difference, to be honest - the SDK isn't small...

jskeet:
Before: 2.21GB  
After: 2.796GB

So yes, it's pretty significant I'm afraid.

ethanbao:
It's fine if that is the minimal installation requirement already. Please add a comment in the Dockerfile about the size increase. Luke once suggested shipping different artman docker images for each languages. We can consider that option if necessary.

Dockerfile
```diff
@@ -138,6 +138,41 @@ RUN apt-get update \
...
+ENV DOTNET_SDK_DOWNLOAD_URL https://dotnetcli.blob.core.windows.net/dotnet/Sdk/$DOTNET_SDK_VERSION/dotnet-dev-ubuntu.16.04-x64.$DOTNET_SDK_VERSION.tar.gz
+
+RUN curl -SL $DOTNET_SDK_DOWNLOAD_URL --output dotnet.tar.gz \
+    && mkdir -p /usr/share/dotnet \
+    && tar -zxf dotnet.tar.gz -C /usr/share/dotnet \
+    && rm dotnet.tar.gz \
+    && ln -s /usr/share/dotnet/dotnet /usr/bin/dotnet
+
+# Trigger the population of the local package cache
+ENV NUGET_XMLDOC_MODE skip
```
ethanbao: Anyway to make nuget work without local package cache?

jskeet: Honestly don't know - this script mostly C&P'd, to be honest.

ethanbao: Unless it will be drastically slower, I would favor the space than time spent on generation. But if it will only save couple of MBs, let's keep it.

jskeet:
Just tried it without the first `dotnet` call, and it's down to 2.495GB, which is a bit more reasonable. Will amend the commit...

jskeet:
Done, PTAL.

PR #240, opened 2017-07-18
Add artman golden test 

ethanbao:
The golden test runs GAPIC generation against the golden input data, then compares the output layout with the expected output. This helps to capture unexpected GAPIC output layout change caused by either by artman, toolkit or other runtime dependency.

Note: a set of static input files (e.g. google/api/http.proto) are being used for the golden test, so it is not going to capture upstream changes to such input.


test/golden_output_test.py
```diff
@@ -0,0 +1,72 @@
...
+Note: a set of static input files (e.g. google/api/http.proto) are being used
+for the golden test, so it is not going to capture upstream changes to such
+input.
+"""
+
+import os
+import subprocess
+
+
+def run_golden_test():
```
geigerj: Why not use unittest?

test/golden_output_test.py
```diff
@@ -0,0 +1,72 @@
...
+    expected = []
+    with open(os.path.join(golden_dir, 'output', 'expected_output.txt')) as f:
+        expected = f.read().splitlines()
+
+    for lang in ['python', 'java', 'ruby', 'nodejs', 'php', 'go', 'csharp']:
+        generate_gapic_library(golden_dir, lang)
+    actual = []
+    for root, subdirs, files in os.walk('/output'):
+        for f in files:
+            actual.append(os.path.join(root, f)[7:])
```
geigerj: `7` is a little magical

test/testdata/golden/gapic/lang/csharp_gapic.yaml
```diff
@@ -0,0 +1,5 @@
+type: com.google.api.codegen.ConfigProto
```
geigerj: Any reason not to use `git submodule`? It may make it easier to update if there is a change in toolkit that requires a change in `googleapis`.

test/golden_output_test.py
ethanbao: True

test/golden_output_test.py
ethanbao: Solely because it technically is not a unit test. And I don't want it to run as part of the regular unit test.

test/testdata/golden/gapic/lang/csharp_gapic.yaml
ethanbao: Originally, I compare the whole output content. Having a static set of input can avoid unnecessary noise caused by the change to googleapis repo itself. Assuming submodule won't update by itself, I can switch to submodule googleapis repo, except that there is no gapic file for library-example at this moment. Anyway to add something into a submodule directory?

test/golden_output_test.py
```diff
@@ -0,0 +1,72 @@
...
+
+    if sorted(expected) != sorted(actual):
```
ethanbao: Switched from sorted() to list.sort(). Per one stackoverflow post, .sort() is faster. Or do I misunderstand your original comment?

geigerj: Ah, the comment was not about performance. My concern was that `actual` is not guaranteed to be in any particular order because `os.walk` does not guarantee an ordering in its return value. Since you didn't order `actual` before you print it, each time you update the baseline `txt` file by copying the printed output, there can be meaningless diffs due to lines being reordered.

test/testdata/golden/gapic/lang/csharp_gapic.yaml
geigerj: Not that I know of. I think we can either

(1) keep the static files checked into `artman`
(2) add GAPIC config for library to `googleapis` and submodule

I suspect (2) will be slightly easier to maintain, but up to you.

test/golden_output_test.py
geigerj: Confusingly, `unittest` is not actually only for unit tests. It actually also runs the `google-cloud-python` system tests. It's not necessary to use it, but it does give you some nice framework bits like assertions and status codes. You should be able to configure this not to run with the regular unit tests in the circleci config, rather than through the choice of framework.

ethanbao: I tried to use unittest. One downsize is that the test failure message is hard to follow (output by unittest.assertListEqual). With current implementation, the message from circleci report is much easier to follow when it fails.

test/testdata/golden/gapic/lang/csharp_gapic.yaml
ethanbao: I tried to submodule the repo and realize that the output_dir and gapic output dir is not consistent. In this test, I am looking for a artman yaml config that is tailed for testing purpose.

test/golden_output_test.py
geigerj: You should be able to customize the error message with `unittest`. But no need to hold this up on the choice of testing framework; we can switch later if desired.

test/testdata/golden/gapic/lang/csharp_gapic.yaml
geigerj: I am still hesitant to check in so many static files for this test, especially since they may need to be maintained as toolkit is updated. I would prefer to check in the test artman yaml to `googleapis` and submodule it, but that doesn't need to be done in this PR.

test/golden_output_test.py
```diff
@@ -0,0 +1,76 @@
...
+
+    output_dir = '/output'
+    actual = []
+    for root, subdirs, files in os.walk(output_dir):
+        for f in files:
+            actual.append(os.path.join(root, f)[len(output_dir):])
+
+    expected.sort()
+    actual.sort()
+    if expected != actual:
```
geigerj: Optional: it would be useful to print out the set differences `expected - actual` and `actual - expected` to summarize what actually differs between actual and expected before you print out the new baseline.

test/golden_output_test.py
lukesneeringer: > Confusingly, `unittest` is not actually only for unit tests. It actually also runs the `google-cloud-python` system tests. 

True, but if artman is using `py.test`, there is no reason to need `unittest.TestCase`. (I can not remember if artman uses `py.test` or not.)

> It's not necessary to use it, but it does give you some nice framework bits like assertions and status codes. You should be able to configure this not to run with the regular unit tests in the circleci config, rather than through the choice of framework.

~in the circleci config~ in nox.

lukesneeringer: I just looked at the code, and we are using `py.test`.

So:

  1. I would ask that you make the module conform to py.test idioms (whether or not you conform to `unittest.TestCase` idioms -- either yes or no is fine with me). In particular that means `test_*.py` files and functions named according to `test_*`. We can use nox to separate unit from system from golden (??) tests.
  2. Similarly `py.test` has _excellent_ stack traces and output and test failure messages, even if you just use `assert` (the keyword). Move to that.
  3. _Do_ separate this from actual unit tests (as you have done).

test/testdata/golden/gapic/lang/csharp_gapic.yaml
lukesneeringer: My general feeling is that if you have a problem, and you use `git submodule` to solve it, you usually have two problems.

This looks a lot like checking the Artman output against a baseline file. That is really an anti-pattern (toolkit notwithstanding). Can we not do that?

test/golden_output_test.py
lukesneeringer: Orthogonal comment: Use `set()`, not `list.sort()`.

geigerj: This is actually the opposite of what I was asking for. I wanted there to be a canonical ordering so that the baseline file will be stable, instead of permitting arbitrary reorderings of the lines of the file between commits.

ethanbao:
Luke, some changes have been made, which unfortunately wiped out your previous comments. Instead of copy paste subset of googleapis repo manually, I submoduled the repo as Jacob suggested. The passed yamls are tailed for this golden test.

Yes, I am comparing the artman output with a baseline fie which lists all the expected output files. Could you elaborate on what do you mean by anti-pattern?

ethanbao:
In the golden test, switch to use set when asserting. The actual output file list is still printed as a list to retain the order from os.walk. I confirmed that the test report (when test fails) looks better than using list. The Java codegen is currently failing for a separate issue that I am currently investigating.

One thing that might be circleci specific is that the git submodule folder won't be available in the artman container running in circleci. To make it work, I add a git clone step in the golden-test circleci config, which is even better than a locked submodule folder, because that helps capture the potential regression caused by upstream common proto/config changes.

geigerj:
> The actual output file list is still printed as a list to retain the order from os.walk
Note that the behavior of `os.walk` with respect to ordering is not defined.

ethanbao:
PTAL. The test is now passing after temporarily changing api_name from "library_example" to "library" in the test artman config. It can be switched back once the toolkit bug gets fixed.

test/test_golden_output.py
```diff
@@ -0,0 +1,67 @@
...
+# limitations under the License.
+
+"""Artman golden tests.
+
+It runs GAPIC generation against the golden input data, can then compare the
+output layout with the expected output. This helps to capture unexpected GAPIC
+output layout change caused by either by artman, toolkit or other runtime
+dependency.
+
+Note: a set of static input files (e.g. google/api/http.proto) are being used
```
lukesneeringer: This _might_ also be an incorrect note based on the discussion about git clone and git submodule above.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "test/testdata/golden/googleapis"]
```
lukesneeringer: This is a little bit weird. In fact, it is weird enough that I am pretty sure it is not what you want. You have _both_ a submodule and an explicit `git clone`. And the way you have them written are not equivalent.

`git submodule` tags a repository as a submodule to the outer repository, and the sub-repository is grafted in _at a specific commit_. That is why if you look farther down in this "files changed" view, you will see this:

screenshot from 2017-07-28 07 59 30

Notice how it says "added at 75c3a5". The repository is grafted in _at a specific point_.
Your `git clone` command, however, (1) plows over the graft, and (2) will clone HEAD. This is going to give you a repository in an inconsistent state. In some ways this is fine; CircleCI is never going to commit the result so it will hit the bit bucket as soon as the VM spins down. On the other hand, this inconsistency is likely to cause a decent bit of heartburn when contrasting local testing with CI as the gulf between 75c3a5 and HEAD widen.

I can not tell from reading this PR _which_ behavior you actually want. If your goal is to always use googleapis pinned to a specific point, then a submodule _might_ be a good answer. If it is (holding off on discussing that until I understand your goals), at minimum your `git clone` above needs to be replaced with `git submodule update --force`.

If your goal is to use googleapis at HEAD, submodules are very much the wrong tool. You would be far better off just `git clone`-ing the repo to a designated location. As I will say below, this will cause problems for local testing, though, which need to be talked through.

lukesneeringer: Okay, I just looked back and saw:

> One thing that might be circleci specific is that the git submodule folder won't be available in the artman container running in circleci. To make it work, I add a git clone step in the golden-test circleci config, which is even better than a locked submodule folder, because that helps capture the potential regression caused by upstream common proto/config changes.

First of all, that is not CircleCI specific. `git clone` does not clone submodules by default. You have to provide `--recurse-submodules` or use `git submodule update` later. (This is an _extremely_ common misunderstanding of how submodules work.)

More importantly, you do not want a submodule. :-) You want to track HEAD. Kill the submodule and check out to a separate location.

The lousy part of this is that it is going to make it somewhat ornery to run the test locally. My recommendation would be to set it up with the following logic:

  1. If `CIRCLECI` is set, check `googleapis` out to a fixed location.
  2. Else, if `GOOGLEAPIS_DIR` is set, skip checkout entirely and just use whatever is there.
  3. Else, `pytest.skip`.

nox.py
```diff
@@ -32,7 +32,8 @@ def unit_tests(session, python_version):
     session.install('-e', '.')
 
     # Run py.test against the unit tests.
-    session.run('py.test', '-rxs', '--cov', '--cov-append', '--cov-report=')
+    session.run('py.test', '-rxs', '--cov', '--cov-append', '--cov-report=',
+                '--ignore=test/test_golden_output.py')
```
ethanbao: Renamed to `golden_test_artman.py` similar to smoketest_artman.py.

test/test_golden_output.py
ethanbao: Good catch. Fixed.

test/test_golden_output.py
```diff
@@ -0,0 +1,67 @@
...
+import os
+import subprocess
+
+
+def test_library_example():
+    golden_dir = os.path.join(
+        os.path.dirname(os.path.realpath(__file__)), 'testdata', 'golden')
+    with open(os.path.join(golden_dir, 'library_example.golden')) as f:
+        expected = set(f.read().splitlines())
+
```
ethanbao: Done.

lukesneeringer:
Reviewing now.

test/golden/artman_golden_test.py
```diff
@@ -0,0 +1,75 @@
...
+"""
+
+import os
+import subprocess
+
+import pytest
+
+def test_library_example(clone_googleapis_for_testing=False):
+    golden_dir = os.path.dirname(os.path.realpath(__file__))
+    if clone_googleapis_for_testing:
```
lukesneeringer: This is a little surprising. It tethers _the clone_ to the test itself, when the clone is actually setup.

What I would have expected would have been for there to be a `--googleapis` argument with _a path_ (rather than a boolean); skip the test if it is not present, and use that path if it is. Then, the `git clone` could have been moved up to `.circleci/config.yml`.

This would allow you, in local testing, to point to an _existing_ clone. There are lots of reasons to want this; pretend, for example, that you want to check against `googleapis-private`, or a non-master branch.

test/golden/artman_golden_test.py
```diff
@@ -0,0 +1,75 @@
...
+        for f in files:
+            actual.append(os.path.join(root, f)[len(output_dir):])
+
+    print('Actual golden output')
+    print('=============================================================')
+    actual.sort()
+    for item in actual:
+        print(item)
+    print('=============================================================')
+    assert expected == set(actual)
```
lukesneeringer: Another problem with this test as written is that it leaves its googleapis clone sitting around. This means (1) the test can not be run twice on the same machine (the `git clone` will error out on a second attempt and therefore `subprocess.check_call` will raise `RuntimeError`) and (2) in a non-CI environment, cruft is left around.

If you do not change the flag as I suggested earlier, at minimum there needs to be a `finally` clause that cleans up what was left behind.

test/golden/artman_golden_test.py
```diff
@@ -0,0 +1,75 @@
...
+    for lang in ['python', 'java', 'ruby', 'nodejs', 'php', 'go', 'csharp']:
+        generate_gapic_library(golden_dir, lang)
+
+    output_dir = '/output'
+    actual = []
+    for root, subdirs, files in os.walk(output_dir):
+        for f in files:
+            actual.append(os.path.join(root, f)[len(output_dir):])
+
+    print('Actual golden output')
```
lukesneeringer: I still wish you would make this an artifact, which would allow it to be tracked over time. I will not block on it though.

test/golden/artman_golden_test.py
```diff
@@ -0,0 +1,75 @@
...
+            'https://github.com/googleapis/googleapis',
+            os.path.join(golden_dir, 'googleapis')
+        ])
+    else:
+        pytest.skip(
+            'Skip the golden test as the --clone_googleapis_for_testing is not '
+            'passed through py.test')
+
+    with open(os.path.join(golden_dir, 'library_example.golden')) as f:
+        expected = set(f.read().splitlines())
```
lukesneeringer: Nitpick: Consider changing this to:

```python
expected = set()
for line in f.read().splitlines():
    line = line.strip()
    if line and not line.startswith('#'):
        expected.add(line)
```

This will have the following effects:

1. Blank lines will be skipped (important if a future committer has an editor that automatically adds blank lines to the end of the file, as many do).

2. Any line starting with `#` will be skipped, allowing comments to exist in the baseline file if needed in the future.

ethanbao:
Address the feedback by

1) Move the git clone into circleci config
2) use `-googleapis_dir` to pass local path to pytest
3) wrote the output to a file and upload it as circleci artifacts. (Proved to work)

PTAL, Luke

ethanbao:
Done :)

ethanbao:
hooray

PR #241, opened 2017-07-18
Move existing artman config as .old in configgen

shinfan:
Previously Artman will not override the existing config file and throw an exception instead.
Change the behavior to move the existing config file as '.old'


artman/tasks/gapic_tasks.py
```diff
@@ -66,14 +66,15 @@ def _move_to(self, gapic_config_dir, gapic_api_yaml):
...
+            olderVersion = conf_out + '.old'
+            print('File already exists, save the old version as ' + olderVersion)
+            self.exec_command(['mv', conf_out, olderVersion])
+        return conf_out
 
     def execute(self, gapic_config_dir, gapic_api_yaml):
         conf_out = self._move_to(gapic_config_dir, gapic_api_yaml)
         self.exec_command(['mkdir', '-p', os.path.dirname(conf_out)])
-        self.exec_command(['mv', gapic_config_dir, conf_out])
+        self.exec_command(['cp', gapic_config_dir, conf_out])
```
ethanbao: Instead of copying the whole dir, can we just copy the specific gapic file?

shinfan: The name was misleading, it is a file not a dir. Changed to gapic_config_path

artman/tasks/gapic_tasks.py
```diff
@@ -66,14 +66,15 @@ def _move_to(self, gapic_config_dir, gapic_api_yaml):
         conf_out = os.path.abspath(gapic_api_yaml[0])
         if os.path.exists(conf_out):
             # TODO (issue #80): no need to test in remote environment
-            raise ValueError(error_fmt + 'File already exists')
-        else:
-            return conf_out
+            olderVersion = conf_out + '.old'
+            print('File already exists, save the old version as ' + olderVersion)
+            self.exec_command(['mv', conf_out, olderVersion])
```
ethanbao: This will fail if running twice. I won't bother to make a perfect change history though.

PR #247, opened 2017-07-27
bump toolkit commit in artman Docker image

ethanbao:
bump toolkit commit in artman Docker image so that the generated GAPICs are compatible with the latest GAX


garrettjonesgoogle:
LGTM

PR #250, opened 2017-08-14
Fix an artman bug that uses default root permission

ethanbao:
Fix an artman bug that uses default root permission for all output when running docker instance.

Note: docker is designed to run in this way.


artman/cli/main.py
```diff
@@ -417,3 +418,23 @@ def _print_log(pipeline_id):
         '    gcloud beta logging read "logName=projects/vkit-pipeline/logs/%s"'
         % pipeline_id,
     )
+
+
+def _chown_for_artman_output():
```
lukesneeringer: Can `g3artman` just set this automatically? It knows who the user and group are.

artman/cli/main.py
```diff
@@ -417,3 +418,23 @@ def _print_log(pipeline_id):
...
+  """Change ownership of artman output if necessary.
+
+  When artman runs in Docker instance, all output files are by default owned by
+  `root`, making it non-editable by Docker host user. When user passes host user
+  id and group id through environment variables via `-e` flag, artman will
```
vchudnov-g: Where is this `-e` flag defined? This makes it sound like it's an artman flag, but I think it's actually a Docker flag. Could you clarify the comment?

artman/cli/main.py
```diff
@@ -417,3 +418,23 @@ def _print_log(pipeline_id):
...
+  change the owner based on the specified user id and group id.
+  """
+  if os.getenv('HOST_USER_ID') and os.getenv('HOST_GROUP_ID'):
```
vchudnov-g: I think it might be cleaner and safer that whatever command starts Docker to run artman either pass the (to `docker`) the `--user` flag or specify the user in the Dockerfile. See https://medium.com/@mccode/understanding-how-uid-and-gid-work-in-docker-containers-c37a01d01cf

One thing I'm not thrilled about with either the current solution in this PR or the Docker-user-setting solution is that the caller can set the environment variables to refer to an arbitrary host or group, not just their own. I think this might be unavoidable with `docker` (which must have the setuid and setgid bits set). So I'd rather use the existing Docker-based solution rather than this.

artman/cli/main.py
ethanbao: It is a docker flag. I will clarify in a follow up PR.

PR #253, opened 2017-08-15
Fix two artman issues

ethanbao:
* Make sure the root directory of the artman output gets its owner permission changed properly as well.
* Fix a php golden test failure.

Note: golden-test is current running against latest artman docker image with a local installed artman python library. The test cannot capture the Dockerfile change that bump the toolkit hash. This needs to be fixed in a follow-up CL.


michaelbausor:
LGTM on the golden test update - this is the expected output once toolkit is updated to the latest version.

PR #254, opened 2017-08-17
Introduce new artman config spec and artman CLI

ethanbao:
The main part of this PR is to introduce a new artman config spec and artman CLI, so that the artman spec is formalized from the artman user perspective. This PR doesn't change the implementation of artman in-memory data model which remains as a dict. This PR parses an artman yaml in the new spec and turn it into a legacy yaml, so that the spec change becomes transparent to artman internal execution engine. A follow-up PR will make another round of refactoring that switches to use the new artman config model (a protobuf class) instead of the legacy config dict.

`artman2` will by default executes inside an artman Docker container, unless `--local` is passed. Once all current artman config yaml switch to the new config spec, `artman2` will be renamed to `artman` and the legacy `artman` will be removed. And `artman` version will become 0.5.*.

To prove the correctness of the conversion, test_converter.py is a unittest to compare the two formats once loaded into artman. In addition to that, artman golden test checks the output using the new and old spec/CLI , and make sure the same set of output gets generated.

This changes make the following changes to the artman spec:
<table>
  <tr>
   <td><strong>Field</strong>
   </td>
   <td><strong>Change</strong>
   </td>
  </tr>
  <tr>
   <td><em>gapic_code_dir</em>
   </td>
   <td>Merge with <em>output_dir</em>
   </td>
  </tr>
  <tr>
   <td><em>output_dir</em>
   </td>
   <td>Remove the config. And use a new logic 1) check artman CLI flag, 2) check
env var, 3) default to a random dir in /tmp
   </td>
  </tr>
  <tr>
   <td><em>packaging, package_type and pipeline</em>
   </td>
   <td>Merge into a new field called <em>type</em> with valid values [GAPIC,
GAPIC_ONLY, GAPIC_CONFIG, GRPC, GRPC_COMMON]
   </td>
  </tr>
  <tr>
   <td><em>import_proto_path</em>
   </td>
   <td>Retained, but become optional
   </td>
  </tr>
  <tr>
   <td><em>gapic_api_yaml</em>
   </td>
   <td>Renamed to <em>gapic_yaml</em>
   </td>
  </tr>
  <tr>
   <td><em>desc_proto_path</em>
   </td>
   <td>Merged with <em>proto_deps</em>
   </td>
  </tr>
  <tr>
   <td><em>enable_batch_generation</em>
   </td>
   <td>Moved into a batch-generation specific config
   </td>
  </tr>
  <tr>
   <td><em>gapic_subcomponent, grpc_subpath, gapic_subpath</em>
   </td>
   <td>Remove because they are no longer needed
   </td>
  </tr>
  <tr>
   <td><em>service_yaml</em>
   </td>
   <td>Type changed from list<string> to string
   </td>
  </tr>
  <tr>
   <td><em>${GOOGLEAPIS} and ${REPO}</em>
   </td>
   <td>Such system variable gets phased out from the user-facing config. For
fields in artman config that point to a file, user can now either specify a path
relative to the dir passed through --output_dir flag or an absolute path (less
common). ${REPO} is effectively replaced by --output_dir flag.
   </td>
  </tr>
</table>

It also make the artman CLI syntax to be:
<table>
  <tr>
   <td><strong>Flag</strong>
   </td>
   <td><strong>Usage</strong>
   </td>
  </tr>
  <tr>
   <td colspan="2" >top-level
   </td>
  </tr>
  <tr>
   <td><code>--config</code>
   </td>
   <td><code>[Optional] Specify path to artman config yaml, which can be either
an absolute path, or a path relative to the current directory. Default to
`./artman.yaml`.</code>
   </td>
  </tr>
  <tr>
   <td><code>--input-dir</code>
   </td>
   <td><code>[Optional] Directory with all input that is needed by artman, which
include but not limited to API protos, service config yaml and GAPIC config
yaml. Default to `.`</code>
   </td>
  </tr>
  <tr>
   <td><code>--output-dir</code>
   </td>
   <td><code>[Optional] Directory to store output generated by artman. Default
to `./artman-genfiles`</code>
   </td>
  </tr>
  <tr>
   <td><code>--verbose, -v</code>
   </td>
   <td><code>[Optional] Show verbose / debug output.</code>
   </td>
  </tr>
  <tr>
   <td><code>--user-config</code>
   </td>
   <td><code>[Optional] User configuration file to stores credentials like
GitHub credentials. Default to `~/.artman/config.yaml`</code>
   </td>
  </tr>
  <tr>
   <td><code>--local</code>
   </td>
   <td><code>[Optional] If specified, running the artman on the local host
machine instead of artman docker instance that have all binaries installed.
Note: one will have to make sure all binaries get installed on the local machine
with this flag, a full list can be found <a
href="https://github.com/googleapis/artman/blob/master/Dockerfile">here</a>.</code>
   </td>
  </tr>
  <tr>
   <td><code>--image</code>
   </td>
   <td><code>[Optional] Specify docker image used by artman when running in a
Docker instance. Each artman release will have default artman docker
image.</code>
   </td>
  </tr>
  <tr>
   <td colspan="2" >`generate` subcommand
   </td>
  </tr>
  <tr>
   <td><code>artifact_name</code>
   </td>
   <td><code>[Required] Name of the artifact for artman to generate. Must match
an artifact in the artman config yaml.</code>
   </td>
  </tr>
  <tr>
   <td colspan="2" >`publish` subcommand
   </td>
  </tr>
  <tr>
   <td><code>artifact_name</code>
   </td>
   <td><code>[Required] Name of the artifact for artman to publish (after
generation). Must match an artifact in the artman config yaml.</code>
   </td>
  </tr>
  <tr>
   <td><code>--target</code>
   </td>
   <td><code>[Required] Specify where the generated artifact should be published
to. It is defined as publishing targets in artman config at artifact
level.</code>
   </td>
  </tr>
  <tr>
   <td><code>--dryrun</code>
   </td>
   <td><code>[Optional] When specified, the publishing step will do everything
except the real remote publish.</code>
   </td>
  </tr>
  <tr>
   <td><code>--github-username</code>
   </td>
   <td><code>[Optional] The GitHub username. Must be set if publishing, but can
come from the user config file.</code>
   </td>
  </tr>
  <tr>
   <td><code>--github-token</code>
   </td>
   <td><code>[Optional] The GitHub personal access token. Must be set if
publishing, but can come from the user config file.</code>
   </td>
  </tr>
</table>


artman/cli/__init__.py
```diff
@@ -16,6 +16,7 @@
 
 from artman.cli import configure
 from artman.cli import main
+from artman.cli import main2
 from artman.cli import support
 
-__all__ = ('configure', 'main', 'support')
+__all__ = ('configure', 'main', 'main2' 'support')
```
lukesneeringer: Comma between `'main2'` and `'support'`. As it stands this evaluates to `'main2support'`.

artman/cli/main2.py
```diff
@@ -0,0 +1,435 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""The new artman CLI with the following syntax.
+
+    artman [Options] generate|publish <artifact_name>
+
+Note: Only local execution is supported as this moment. The CLI syntax is beta,
```
lukesneeringer: Use Restructured Text for notes:

```rst
.. note::
    Only local execution is supported at this moment. The CLI syntax is
    beta, and might have changes in the future.
```

artman/cli/main2.py
```diff
@@ -0,0 +1,435 @@
...
+from artman.utils import config_util
+from artman.utils.logger import logger, setup_logging
+
+ARTMAN_DOCKER_IMAGE = 'googleapis/artman:0.4.12'
+
+
+def main(*args):
+    """Main method of artman.
+
+    Note: artman currently won't work if the input directory doesn't contain
```
lukesneeringer: Move this out of the docstring and into a regular `#` comment.

artman/cli/main2.py
```diff
@@ -0,0 +1,435 @@
...
+        help='[Optional] The GitHub username. Must be set if publishing the '
+        'artifact to github, but can come from the user config file.', )
+    parser_publish.add_argument(
+        '--github-token',
+        default=None,
+        help='[Optional] The GitHub personal access token. Must be set if '
+        'publishing the artifact to github, but can come from the user '
+        'config file.', )
+    parser_publish.add_argument(
+        '--dryrun',
```
lukesneeringer: Spell this `--dry-run`.

artman/cli/main2.py
```diff
@@ -0,0 +1,435 @@
...
+        artifact_config = loader.load_artifact_config(
+            artman_config_path, flags.artifact_name, flags.input_dir)
+    except ValueError as ve:
+        logger.error('Artifact config loading failed with `%s`' % ve)
+
+    # If we were given just an API or BATCH, then expand it into the --config
+    # syntax.
+    shared_config_name = 'common.yaml'
+    if artifact_config.language == config_pb2.Artifact.RUBY:
+        # TODO(ethanbao): Figure out why.
```
lukesneeringer: I am guessing this is probably because Ruby has an additional step to generate the bogus alternate pb2s so that yard can generate documentation for them.

In a perfect world, I would like to use protoc-docs-plugin to solve this, but the Ruby protobuf output lacks insertion points.

artman/cli/main2.py
```diff
@@ -0,0 +1,435 @@
...
+        shared_config_name = 'doc.yaml'
+
+    legacy_config_dict = converter.convert_to_legacy_config_dict(
+        artifact_config, flags.input_dir, flags.output_dir)
+    logger.debug('Below is the legacy config after conversion:\n%s' %
+                 pprint.pformat(legacy_config_dict))
+    tmp_legacy_config_yaml = '%s.tmp' % artman_config_path
+    with open(tmp_legacy_config_yaml, 'w') as outfile:
```
lukesneeringer: Minor preference for `io.open`; it will make unit testing this line easier.

artman/cli/main2.py
```diff
@@ -0,0 +1,435 @@
...
+    config = ','.join([
+        '{artman_config_path}',
+        '{googleapis}/gapic/lang/{shared_config_name}',
+    ]).format(
+        artman_config_path=tmp_legacy_config_yaml,
+        googleapis=googleapis,
+        shared_config_name=shared_config_name, )
+
+    # Set the pipeline as well as package_type and packaging
+    artifact_type = artifact_config.type
```
lukesneeringer: I would find this huge switch statement easier to reason about if multiple items with the same result were grouped together:

```python
Artifact = config_pb2.Artifact
if artifact_type in (Artifact.GAPIC, Artifact.GAPIC_ONLY):
    pipeline_name = 'GapicClientPipeline'
elif artifact_type in (Artifact.GRPC, Artifact.GRPC_COMMON):
    pipeline_name = 'GrpcClientPipeline'
elif artifact_type == Artifact.GAPIC_CONFIG:
    pipeline_name = 'GapicConfigPipeline'
else:
    raise ValueError('Unrecognized artifact.')
```

artman/cli/main2.py
lukesneeringer: And the target variable `dry_run`.

artman/cli/main2.py
```diff
@@ -0,0 +1,435 @@
...
+        if publishing_config.type == config_pb2.Artifact.PublishTarget.GITHUB:
+            pipeline_args['publish'] = 'local' if flags.dryrun else 'github'
+            pipeline_args['github'] = support.parse_github_credentials(
+                argv_flags=flags,
+                config=user_config.get('github', {}), )
+            repos = pipeline_args.pop('git_repos')
+            pipeline_args['git_repo'] = support.select_git_repo(
+                repos, publishing_config.name)
+        else:
+            raise NameError(
```
lukesneeringer: `ValueError` or `RuntimeError`. There is a very specific meaning for `NameError` (use of an undefined variable) which is not correct here.

lukesneeringer: Alternatively, define your own exception that subclasses one of those. :-)

artman/cli/support.py
```diff
@@ -146,3 +145,27 @@ def select_git_repo(git_repos, target_repo):
...
+    """Checks whether all docker-related components have been installed."""
+
+    try:
+        subprocess.check_output(['docker', '--version'])
+    except OSError:
+        logger.error(
+            'Docker not found on path or is not installed. Refert to '
+            'https://docs.docker.com/engine/installation about how to install '
+            'Docker on your local machine.')
+        sys.exit(1)
```
lukesneeringer: Nitpick: Use something other than `1` as the exit code here, since you will likely want to bubble the exit code from gradlew when you run it, and that is likely to often be `1`.

artman/config/__init__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2017 Google
```
lukesneeringer: You can skip the license for blank files.

artman/config/converter.py
```diff
@@ -0,0 +1,160 @@
...
+    elif language == 'csharp':
+        return 'csharp/google-cloud-%s' % api_name
+    elif language == 'go':
+        return 'gapi-cloud-%s-go' % api_name
+    elif language == 'nodejs':
+        return 'js/%s-%s' % (api_name, api_version)
+    elif language == 'php':
+        return 'google-php-cloud-%s-%s' % (api_name, api_version)
+    elif language == 'python':
+        return 'gapic-google-cloud-%s-%s' % (api_name, api_version)
```
lukesneeringer: Unrelated, but can this be `python/{name}_{version}`? Thanks.

artman/cli/main2.py
geigerj: Confirming what Luke said. Ethan and I have already chatted offline about this; I think it makes sense to leave this here in the meantime.

artman/cli/main2.py
```diff
@@ -0,0 +1,435 @@
...
+def parse_args(*args):
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '--config',
+        type=str,
+        default='artman.yaml',
+        help='[Optional] Specify path to artman config yaml, which can be '
+        'either an absolute path, or a path relative to the input '
+        'directory (specified by `--input-dir` flag). Default to '
+        '`artman.yaml`', )
```
geigerj: Is it meaningful to have a default, or should we require this?

artman/config/loader.py
```diff
@@ -0,0 +1,137 @@
...
+    valid_values = []
+    for artifact in artman_config.artifacts:
+        valid_values.append(artifact.name)
+        if artifact.name == artifact_name:
+            artifact_config.MergeFrom(artifact)
+            _validate_artifact_config(artifact_config)
+            return _normalize_artifact_config(artifact_config, input_dir)
+
+    raise ValueError(
+        'No artifact with `%s` configured in artman yaml. Valie values are %s'
```
geigerj: sp(Valie)

artman/config/loader.py
```diff
@@ -0,0 +1,137 @@
...
+    if not os.path.isabs(artifact_config.service_yaml):
+        artifact_config.service_yaml = os.path.join(
+            input_dir, artifact_config.service_yaml)
+
+    if not os.path.isabs(artifact_config.gapic_yaml):
+        artifact_config.gapic_yaml = os.path.join(
+            input_dir, artifact_config.gapic_yaml)
+
+    if not artifact_config.import_proto_path:
+        artifact_config.import_proto_path[:] = [input_dir]
```
geigerj: Why is slice assignment needed here?

artman/config/loader.py
```diff
@@ -0,0 +1,137 @@
...
+def _normalize_artifact_config(artifact_config, input_dir):
+    """Normalize the config protobuf based on flags passed from command line.
+
+    Note: we are not normalizing output folders because they are no longer
+    configurable, and the output folder calculation now happens in converter.
+    Once the individual GAPIC output folder becomes configurable, that folder
+    name calculation logic should be moved from converter into this method.
+    """
+    # Normalize the input file or folder by prefixing with input_dir if needed.
+    if not os.path.isabs(artifact_config.service_yaml):
```
geigerj: Suggest defining a helper to remove some of this repetition:

```
def normalize(path, prefix):
  return path if os.path.isabs(path) else os.path.join(prefix, path)
```

artman/config/proto/config_pb2.py
```diff
@@ -0,0 +1,569 @@
+# Generated by the protocol buffer compiler.  DO NOT EDIT!
```
geigerj: Have you thought about having artman generate this on-the-fly instead of checking it in?

artman/tasks/publish/github.py
```diff
@@ -143,6 +143,7 @@ def execute(self, git_repo, api_name, api_version, language, github,
             self.exec_command(['rm', '-rf', repo_temp_dir])
 
             # Change the working directory back to where we started.
+            print(original_directory)
```
geigerj: Remove this line? (Or use `logger` if it should be logged)

artman/cli/main2.py
```diff
@@ -0,0 +1,435 @@
...
+            index = line.index(':')
+            line = line[:index + 2] + '<< REDACTED >>'
+        logger.info('  {0}'.format(line))
+
+    # Clean up the tmp legacy artman config.
+    os.remove(tmp_legacy_config_yaml)
+
+    # Return the final arguments.
+    # This includes a pipeline to run, arguments, and whether to run remotely.
+    return (pipeline_name, pipeline_args)
```
geigerj: nit: unnecessary parens

artman/cli/main2.py
```diff
@@ -0,0 +1,435 @@
...
+    if not os.path.isfile(artman_config_path):
+        logger.error(
+            'Artman config file `%s` doesn\'t exist.' % artman_config_path)
+        sys.exit(96)
+
+    try:
+        artifact_config = loader.load_artifact_config(
+            artman_config_path, flags.artifact_name, flags.input_dir)
+    except ValueError as ve:
+        logger.error('Artifact config loading failed with `%s`' % ve)
```
geigerj: Why not `sys.exit`? Isn't invalid config fatal?

artman/cli/main2.py
```diff
@@ -0,0 +1,435 @@
...
+
+    # If we were given just an API or BATCH, then expand it into the --config
+    # syntax.
+    shared_config_name = 'common.yaml'
+    if artifact_config.language == config_pb2.Artifact.RUBY:
+        # TODO(ethanbao): Figure out why.
+        shared_config_name = 'doc.yaml'
+
+    legacy_config_dict = converter.convert_to_legacy_config_dict(
```
geigerj: At the point that you have the `legacy_config_dict`, can't we reuse the code `main.py`? Why do we need the rest of the `main2.py` file at all?

artman/cli/__init__.py
ethanbao: My bad. It surprised me that nothing fail with this.

artman/cli/main2.py
ethanbao: Done

artman/cli/main2.py
ethanbao: Done

artman/cli/main2.py
ethanbao: Done

artman/cli/main2.py
ethanbao: Yes. Discussed with Jacob offline, and leave this as-is.

artman/cli/main2.py
ethanbao: Done.

artman/cli/support.py
ethanbao: You used `64`. Would you recommend I use a different error code or simply use 64?

artman/cli/main2.py
ethanbao: I actually changed to log the error here and exit.

artman/cli/support.py
ethanbao: Used `128` for now.

artman/config/converter.py
ethanbao: Made both new __init__.py files empty.

artman/cli/main2.py
ethanbao: I have mixed feeling. I don't like such magic for any input file, but it is also not too controversial if you see it as BUILD file of bazel, setup.py for python, or pom.xml for maven.

artman/cli/main2.py
ethanbao: Yes. Changed to sys.exit.

artman/cli/main2.py
ethanbao: The only reason is that I don't bother to have a refactoring to main.py in this PR, as the current main.py will be removed in a follow-up PR and be replaced by this main2.py (I will rename).

artman/cli/main2.py
ethanbao: Done. Also removed the comment for the remote execution part.

artman/config/loader.py
ethanbao: :)

artman/config/proto/config_pb2.py
ethanbao: I know we did this for toolkit in Java with a gradle plugin. Do you happen to know whether protobuf python has such feature or not?

artman/tasks/publish/github.py
ethanbao: My bad. It was not intended to be checked in. Removed.

artman/config/proto/config_pb2.py
geigerj: I'm not sure about the dynamically loading a `.proto` in the Python proto runtime. I was thinking it would just involve a step of artman that executes `protoc`.

artman/cli/main2.py
geigerj: Ack. As long as this is done in a follow-up so that these files don't diverge.

artman/cli/main2.py
geigerj: Ack.

artman/cli/main2.py
```diff
@@ -0,0 +1,435 @@
...
+            container as local googleapis directory.
+    Returns:
+        The output directory with artman-generated files.
+    """
+    ARTMAN_CONTAINER_NAME = 'artman-docker'
+    input_dir = flags.input_dir
+    output_dir = flags.output_dir
+    artman_config_dirname = os.path.dirname(flags.config)
+    user_config = os.path.join(os.path.expanduser('~'), '.artman')
+    docker_image = 'e57acef1a4b1'  # flags.image
```
geigerj: `... = flags.image or 'e57acef1a4b1'`?

ethanbao: Good catch. Updated.

artman/cli/main2.py
lukesneeringer: This is just a minor readability item; it would be easier to understand (for me) if each final destination had one predicate, instead of having three destinations and five predicates.

artman/config/proto/config_pb2.py
lukesneeringer: It would not be very hard to make the distribution step run protoc.

artman/cli/main2.py
ethanbao: Ack. My bad. You gave example already.

artman/config/proto/config_pb2.py
ethanbao: That was my original instinct. I failed to find a doc or example on how to do that. Can you give me some guidance? The official python protobuf doc (https://developers.google.com/protocol-buffers/docs/pythontutorial) has very limited information regarding the distribution (proto vs pb2).

ethanbao:
Thank you all for the review. For the protoc part, I will follow up with another PR once I figure out how to do that with python distribution.

PR #255, opened 2017-08-18
Support proto generated package GA version

shinfan:
Config change PR: googleapis/googleapis#427


shinfan:
FYI the smoke tests failed due to a change is made to the api_defaults config, which should be fixed after the PR is merged.

PR #256, opened 2017-08-21
Introduce artman/cli/converter.py that converts legacy artman config into new one

ethanbao:
This basically does the reverse of artman/config/converter.py, and can be removed after all current artman configs get upgraded to the new syntax. test/smoketest_artman.py (not part of CI) is now testing with artman2, which is currently failing until https://github.com/googleapis/googleapis/pull/429 gets merged. Plus, this PR also fix two minor issues in artman/config/converter.py

Note: there is no test code for this new converter, as you can treat googleapis PR#429 as its test cases.


test/smoketest_artman.py
```diff
@@ -101,10 +100,15 @@ def parse_args(*args):
              'repo in order for smoketest to run properly. If not specified, '
              'all APIs will be tested. APIs in the blacklist will not be '
              'tested.')
+    parser.add_argument(
+        '--input-dir',
+        default='/googleapis',
```
eoogbe: Should this be `~/googleapis`?

artman/cli/converter.py
```diff
@@ -0,0 +1,306 @@
...
+
+
+def parse_args(*args):
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        'config',
+        type=str,
+        help='Path to the legacy artman config')
+    parser.add_argument(
+        '--output',
```
eoogbe: If we drop this arg and always print to stdout, we can collect the output in a file using `>`.

artman/cli/converter.py
```diff
@@ -0,0 +1,306 @@
...
+
+
+def _compute_src_proto_paths(src_proto_paths):
+    result = []
+    for src_proto_path in src_proto_paths:
+        result.append(_sanitize_repl_var(src_proto_path))
+    return result
+
+
+def _sanitize_repl_var(value):
```
eoogbe: Should this return the unchanged value at the end?

artman/cli/converter.py
```diff
@@ -0,0 +1,306 @@
...
+
+            # Compute publishing targets
+            if 'git_repos' in legacy_artifact_config:
+                artifact.publish_targets.extend(_compute_publish_targets(
+                    legacy_artifact_config.get('git_repos'), common_git_repos))
+            result.append(artifact)
+    return result
+
+
+def _compute_package_version(legacy_package_version):
```
eoogbe: This appears to be unused.

artman/cli/converter.py
```diff
@@ -0,0 +1,306 @@
...
+
+
+def _sanitize_repl_var(value):
+    if value.startswith('${GOOGLEAPIS}/'):
+        return value.replace('${GOOGLEAPIS}/', '')
+    if value.startswith('${REPOROOT}/'):
+        return value.replace('${REPOROOT}/', '')
+
+
+def camel_to_underscore(name):
```
eoogbe: Why is this public? Same with `parse_args`, `convert_json`, and `represent_ordereddict`.

artman/cli/converter.py
```diff
@@ -0,0 +1,306 @@
...
+                new_d[camel_to_underscore(k)] = result
+            else:
+                new_d[camel_to_underscore(k)] = v
+        else:
+            new_d[camel_to_underscore(k)] = v
+    return new_d
+
+
+def _write_pb_to_yaml(pb, output):
+    # Add yaml representer so taht yaml dump can dump OrderedDict. The code
```
eoogbe: s/taht/that

test/smoketest_artman.py
```diff
@@ -38,58 +38,57 @@
 SUPPORTED_LANGS = ['python', 'java', 'ruby', 'nodejs', 'php', 'go', 'csharp']
```
eoogbe: Can there be one place to go to that lists all the languages? This list seems to get repeated quite often.

artman/cli/converter.py
ethanbao: That is true. Different tooling does different flavor for the flag. My convertor/reformatter will support `-i` or `--in-place`. I am not too worried about this specific flag, as I am very likely be the only person who will need this tool. I will probably remove this tooling next quarter anyway.

artman/cli/converter.py
ethanbao: Hmmm. Then there won't be sanitization at all, right?

artman/cli/converter.py
ethanbao: Changed to private.

artman/cli/converter.py
ethanbao: Good catch. Removed.

test/smoketest_artman.py
ethanbao: This smoketests is supposed to be running inside artman docker container in most cases, where /googleapis is where the input will locate by default.

artman/cli/converter.py
ethanbao: Done

artman/cli/converter.py
eoogbe: IIUC the return value is used, but the function doesn't always return a value. If no sanitization is needed, the unchanged value should be returned.

ethanbao: Oh. Misunderstood your comment. You are right. Fixed.

test/smoketest_artman.py
eoogbe: Ack.

artman/cli/converter.py
```diff
@@ -0,0 +1,297 @@
...
+    elif legacy_common.get('package_type', '') == 'grpc_common':
+        return Artifact.GRPC_COMMON
+
+    return Artifact.GAPIC
+
+
+def _compute_publish_targets(git_repos, common_git_repos):
+    result = []
+    for k, v in git_repos.items():
+        location = (common_git_repos[k]['location']
```
geigerj: style nit: use the standard multi-line `if` statement when the ternary operator doesn't fit in one line

artman/cli/converter.py
```diff
@@ -0,0 +1,297 @@
...
+                    'generated_package_version')
+                if 'lower' in legacy_package_version:
+                    artifact.package_version.grpc_dep_lower_bound = (
+                        legacy_package_version.get('lower'))
+                if 'upper' in legacy_package_version:
+                    artifact.package_version.grpc_dep_upper_bound = (
+                        legacy_package_version.get('upper'))
+
+            # Compute publishing targets
+            if 'git_repos' in legacy_artifact_config:
```
geigerj: If a lang doesn't define `git_repos`, doesn't it still automatically inherit the common git repo? (Maybe I am mistaken about this bit of existing artman logic, but it seems like all langs should stage to `api-client-staging` by default.) This looks like the new artifact will have no publish target if the legacy lang config doesn't define a git repo.

artman/cli/converter.py
```diff
@@ -0,0 +1,297 @@
...
+            if 'artifact' in path:
+                mapping.name = path.get('artifact')
+            result.append(mapping)
+    return result
+
+
+def _compute_deps(proto_deps):
+    result = []
+    for dep in proto_deps:
+        new_dep = Artifact.ProtoDependency()
```
geigerj: nit: `Artifact.ProtoDependency(name=dep)`

here and throughout

artman/cli/converter.py
```diff
@@ -0,0 +1,297 @@
...
+
+    return Artifact.GAPIC
+
+
+def _compute_publish_targets(git_repos, common_git_repos):
+    result = []
+    for k, v in git_repos.items():
+        location = (common_git_repos[k]['location']
+                    if k in common_git_repos.keys() else v['location'])
+        target = Artifact.PublishTarget()
```
geigerj: nit: `Artifact.PublishTarget(name=k, location=location, type=Artifact.PublishTarget.GITHUB)`

here and throughout

artman/cli/converter.py
lukesneeringer: Actually, use the standard multi-line `if` statement always.

artman/cli/converter.py
```diff
@@ -0,0 +1,306 @@
...
+# limitations under the License.
+
+"""Artman config converter.
+
+It converts the legacy artman config file into the new format. This CLI can be
+removed after all current artman users migrate to the new artman config format.
+"""
+
+from __future__ import absolute_import
+import argparse
```
lukesneeringer: Also:

  * Empty line after `from __future__ import absolute import`.
  * Empty line before `import yaml`, which is not in the standard library.

artman/cli/converter.py
```diff
@@ -0,0 +1,297 @@
...
+import io
+import json
+import re
+import sys
+import os
+import yaml
+
+
+from collections import OrderedDict
```
lukesneeringer: This belongs with the stdlib stuff.

artman/cli/converter.py
```diff
@@ -0,0 +1,306 @@
...
+
+from __future__ import absolute_import
+import argparse
+import io
+import json
+import re
+import sys
+import os
+import yaml
+
```
lukesneeringer: Note to self: Add a linter to artman. :-)

artman/cli/converter.py
lukesneeringer: Also add `from __future__ import print_function`, which is a prerequisite for @geigerj's comment [here](https://github.com/googleapis/artman/pull/256/files#r134620455) to not be an error in Python 2.

artman/cli/converter.py
lukesneeringer: The tool in its existing form is not careful about what goes to `sys.stdout` vs. `sys.stderr`, which would be a prerequisite to using redirection.

artman/cli/converter.py
```diff
@@ -0,0 +1,297 @@
...
+        if lang not in legacy_config:
+            continue
+        legacy_artifact_config = legacy_config[lang]
+        if legacy_artifact_config:
+            artifact = Artifact()
+            # Name the artifact as `{lang}_gapic`.
+            artifact.name = '%s_gapic' % lang
+            artifact.language = Artifact.Language.Value(lang.upper())
+
+            if 'release_level' in legacy_artifact_config:
```
lukesneeringer: What will this default to if it is unset?

artman/cli/converter.py
lukesneeringer: Yes, it inherited in the former structure.

lukesneeringer: And it is a deep merge. `dict.extend` will not work.

artman/cli/converter.py
```diff
@@ -0,0 +1,297 @@
...
+        common_git_repos = legacy_common.get('git_repos')
+    else:
+        common_git_repos = []
+
+    for lang in LANGS:
+        if lang not in legacy_config:
+            continue
+        legacy_artifact_config = legacy_config[lang]
+        if legacy_artifact_config:
+            artifact = Artifact()
```
lukesneeringer: Although it is not enforced, protobuf objects are supposed to not be mutated. Prefer:

```python
Artifact(
    language=lang.upper(),
    name='%s_gapic' % lang,
)
```

(I _think_ you can omit `Artifact.Language.Value` at that point, but keep it if you get any kind of error. ;-))

artman/cli/converter.py
```diff
@@ -0,0 +1,297 @@
...
+            target.directory_mappings.extend(
+                _compute_directory_mappings(v['paths']))
+        result.append(target)
+    return result
+
+
+def _compute_directory_mappings(paths):
+    result = []
+    for path in paths:
+        mapping = Artifact.PublishTarget.DirectoryMapping()
```
lukesneeringer: In general in places like this, you should compute a dictionary first and then assign it to the pb2 at the end.

artman/cli/converter.py
```diff
@@ -0,0 +1,297 @@
...
+
+def _compute_src_proto_paths(src_proto_paths):
+    result = []
+    for src_proto_path in src_proto_paths:
+        result.append(_sanitize_repl_var(src_proto_path))
+    return result
+
+
+def _sanitize_repl_var(value):
+    if value.startswith('${GOOGLEAPIS}/'):
```
lukesneeringer: Why is this `.startswith`? Do you only want this behavior if it is at the beginning, as opposed to in the middle or end of a string? (Honestly unsure.)

artman/cli/converter.py
```diff
@@ -0,0 +1,297 @@
...
+        return value.replace('${GOOGLEAPIS}/', '')
+    if value.startswith('${REPOROOT}/'):
+        return value.replace('${REPOROOT}/', '')
+    return value
+
+def _camel_to_underscore(name):
+    camel_pat = re.compile(r'([A-Z])')
```
lukesneeringer: You can just use `re.sub` here. You are compiling every time so you actually _lose_ efficiency.

artman/cli/converter.py
```diff
@@ -0,0 +1,297 @@
...
+    return camel_pat.sub(lambda x: '_' + x.group(1).lower(), name)
+
+
+def _convert_json(d):
+    """Convert the dict to turn all key into lower underscore case."""
```
lukesneeringer: Nitpick: We call it "snake case" or "snake_case".

lukesneeringer: Also, not a huge deal, but this is the kind of thing I tend to prefer third-party packages for (e.g. https://pypi.org/project/stringcase/)

artman/cli/converter.py
geigerj: Luke, my comment comes from the Google style guide: https://google.github.io/styleguide/pyguide.html?showone=Conditional_Expressions#Conditional_Expressions

artman/cli/converter.py
ethanbao: In my search, I don't see any existing case that put "${GOOGLEAPIS}" in any place other than the beginning. Because the new spec gets rid of that, no one will ever run into this. But this indeed is an unnecessary checking.

artman/cli/converter.py
ethanbao: Actually, I am not that familiar with python protobuf API. In Java, you will have Builder class. What is the canonical way of doing that in python protobuf?

geigerj: @ethanbao In Python, you can use optional parameters in the constructor, as Luke and I suggested.

artman/cli/converter.py
ethanbao: Did what Jacob suggested, except that instead of do sys.stderr and then sys.exit, I simply put do sys.exit(error_msg).

artman/cli/converter.py
ethanbao: Done. I guess that is why python protobuf doesn't need builder pattern.

artman/cli/converter.py
geigerj: Not that I'm aware of. AFAIK builders are not typically used with languages that support optional params.

ethanbao: That makes sense.

artman/cli/converter.py
lukesneeringer: Ack.

artman/cli/converter.py
lukesneeringer: `sys.exit` takes _an int_, not a string.

lukesneeringer: Actually, [it can take a string too](https://docs.python.org/3/library/sys.html#sys.exit), for exactly @ethanbao's use case. TIL!

artman/cli/converter.py
```diff
@@ -39,13 +41,17 @@ def main(*args):
...
-    legacy_config = _load_legacy_config_dict(os.path.abspath(flags.config))
-    new_config = _convert(legacy_config)
-    _write_pb_to_yaml(new_config, flags.output)
+    try:
+        flags = _parse_args(*args)
+        legacy_config = _load_legacy_config_dict(os.path.abspath(flags.config))
+        new_config = _convert(legacy_config)
+        _write_pb_to_yaml(new_config, flags.output)
+    except Exception as e:
+        traceback.print_exc(file=sys.stdout)
```
geigerj: nit: typically this would go to stderr, I think

PR #261, opened 2017-09-01
Convert Python output into a single, discrete package.

lukesneeringer:
This PR converts the Python output into a discrete package, rolling in the gRPC/proto and GAPIC output together. It does this intrinsically, and does not rely on artman config.

*This change necessitates two other changes.*

  1. toolkit must be updated to import the protos from the new path (updated to match the gapic path) of `google.cloud.api_vN.proto`. @landrito intends to do this.
  2. All references to `artifact: grpc` must be removed from all artman configs.


lukesneeringer:
@ethanbao This change _will_ break your golden test. Since I can not run it locally, I am assuming that you will make the appropriate updates there. Edits to the PR are allowed.

landrito:
The necessary toolkit change is here:
https://github.com/googleapis/toolkit/pull/1537.

artman/tasks/python_grpc_tasks.py
```diff
@@ -94,35 +94,41 @@ def _extract_base_dirs(self, proto_file):
...
         Works with arbitrary separator (e.g., '/' for import statements,
         '.' for proto package statements, os.path.sep for filenames)
         """
-        # Skip common protos
-        pkg_list = pkg.split(sep)
-
-        dotted = '.'.join(pkg_list)
+        if sep != '.' and pkg.endswith('.proto'):
+            dotted = pkg[:-6].replace(sep, '.')
+            suffix = '.proto'
```
ethanbao: Nit: Move suffix before 'dotted' assignement so that you can use len(suffix) instead of 6.

artman/tasks/python_grpc_tasks.py
```diff
@@ -94,35 +94,41 @@ def _extract_base_dirs(self, proto_file):
...
+            dotted = pkg.replace(sep, '.')
+            suffix = ''
+
+        # Sanity check: Do not transform common protos.
         for common_pkg in common_protos:
             if dotted.startswith(common_pkg):
-                return sep.join(pkg_list)
+                return pkg
+
+        # Transform into the ideal proto path.
```
ethanbao: I don't know the knowledge about why this transforming is needed. But if this is going to work when people do `pip install`, I am okay with any transforming.

test/test_python_package_change_task.py
```diff
@@ -49,20 +49,25 @@ def test__extract_base_dirs(self):
 
     def test__transfom(self):
         # Simple package transformations with arbitrary separator
-        self.assertEqual(self._TASK._transform('google.service', '.', []),
-                         'google.cloud.proto.service')
-        self.assertEqual(self._TASK._transform('google/other', '/', []),
-                         'google/cloud/proto/other')
+        self.assertEqual(self._TASK._transform('google.service.v1', '.', []),
+                         'google.service_v1.proto')
```
ethanbao: Any reason you removed the `cloud` from the path?

lukesneeringer: `_transform` no longer _adds_ cloud if it is not already there. I sanity checked with Pub/Sub, Spanner, and Vision and this seems to consistently do the right thing.

artman/tasks/python_grpc_tasks.py
```diff
@@ -94,35 +94,41 @@ def _extract_base_dirs(self, proto_file):
             return ''
 
     def _transform(self, pkg, sep, common_protos):
-        """Add 'proto' package after 'google' or 'google.cloud'
+        """Transform to the appropriate proto package layout.
```
geigerj: nit: Would be helpful to describe what "appropriate" means in this context.

artman/tasks/python_grpc_tasks.py
```diff
@@ -94,35 +94,41 @@ def _extract_base_dirs(self, proto_file):
...
+        # Sanity check: Do not transform common protos.
         for common_pkg in common_protos:
             if dotted.startswith(common_pkg):
-                return sep.join(pkg_list)
+                return pkg
+
+        # Transform into the ideal proto path.
+        # What essentially should happen here is that "{api}.{vN}" should
+        # change to "{api}_{vN}.proto".
+        dotted = re.sub(r'\.v([\da-z_]*)([\d]+)\b', r'_v\1\2.proto', dotted)
```
geigerj: We have internal customers who use `v1alpha`, which I think this doesn't match (see: https://github.com/googleapis/toolkit/issues/1335). Ideally this should mirror the version matcher used in api-compiler: https://github.com/googleapis/api-compiler/blob/master/src/main/java/com/google/api/tools/framework/aspects/versioning/model/ApiVersionUtil.java#L40

Also, you really only want to replace the rightmost occurrence of this match, but maybe we don't need to worry about that given how rare it should be to have any other package match the version regex.

artman/tasks/python_grpc_tasks.py
```diff
@@ -145,3 +151,59 @@ def _copy_and_transform_directories(
                     self.exec_command(['mkdir', '-p', sub_new_src])
                 self._copy_proto(
                     proto, os.path.join(sub_new_src, dest), common_protos)
+
+
+class PythonMoveProtosTask(task_base.TaskBase):
```
geigerj: Why is this done here, instead of when protoc is run? (I.e., can we just have protoc write to the right directory to begin with?)

artman/tasks/python_grpc_tasks.py
landrito: It would be good to note that the apiVersionUtil regex has not been updated to use the point release syntax.

lukesneeringer: Good catch! (I agree that we do not need to worry about the right-most issue.)

artman/tasks/python_grpc_tasks.py
lukesneeringer: I will check on this.

artman/tasks/python_grpc_tasks.py
lukesneeringer: Updated in 421db76.

lukesneeringer:
> Would still prefer a general-purpose regex like api-compiler's to an edge case, but I'm not going to block approval on it, since we only have one real-world example where this is problematic.

I might be lacking in context here. I do not really know what api-compiler _is_. I would be happy to borrow its prior art. :-)

(I do also think that sometimes, two clear regexes can be preferable to one unclear one. That may or may not be applicable here.)

PR #263, opened 2017-09-06
Using versioned common configs/protos under {googleapis}/[gapic|proto] when running artman inside Docker container

ethanbao:
Currently, the codegen has coupling with some shared configuration yaml under those directories, causing library generation to fail when a breaking change is made to such shared configuration file. This delivers a poor user experience to artman users, as their library generation could fail without any change at API proto side.

This logic can be removed after https://github.com/googleapis/toolkit/issues/1450 is fixed at toolkit side. Since that might take a while, making this change at artman side.


Dockerfile
```diff
@@ -196,4 +199,7 @@ RUN mkdir -p /root/
 ADD artman-user-config-in-docker.yaml /root/.artman/config.yaml
 
 # Install artman.
-RUN pip3 install googleapis-artman==0.4.13
+# RUN pip3 install googleapis-artman==0.4.13
```
geigerj: Do you mean to remove this comment?

artman/cli/main2.py
```diff
@@ -74,6 +76,36 @@ def main(*args):
...
+        DIRS = [
+            'gapic/core',
+            'gapic/lang',
+            'gapic/packaging'
+        ]
+        for src_dir in DIRS:
+            # /googleapis is the root of the versioned googleapis repo
+            # inside Artman Docker image.
+            abs_src_dir = os.path.join('/googleapis', src_dir)
+            for root, dirs, files in os.walk(abs_src_dir):
```
geigerj: Don't you only want to copy `googleapis/gapic`? IIUC with the current setup you'll override any updates to the protos that the user has made if they're running in docker.

Dockerfile
ethanbao: My bad. Uncommented.

artman/cli/main2.py
ethanbao: The src_dir contains "gapic" in the path (see DIRS variable). Technically, I only need three subdirectories under gapic (i.e. core, lang and packaging).

geigerj: Ack, sorry, misread the code.

artman/cli/main2.py
```diff
@@ -74,6 +76,36 @@ def main(*args):
...
+    # coupling with some shared configuration yaml under those directories,
+    # causing library generation to fail when a breaking change is made to such
+    # shared configuration file. This delivers a poor user experience to artman
+    # users, as their library generation could fail without any change at API
+    # proto side.
+    # TODO(ethanbao): Remove this logic once
+    # https://github.com/googleapis/toolkit/issues/1450 is fixed.
+    if os.getenv(RUNNING_IN_ARTMAN_DOCKER_TOKEN):
+        # Only doing this when running inside Docker container
+        DIRS = [
```
geigerj: This is not a module constant, so rename to snake_case (`common_config_dirs`?)

ethanbao: Done.

PR #270, opened 2017-09-14
Do not let artman send duplicate protos to protoc.

lukesneeringer:
It turns out `find_protos` can return duplicate results. Good times.
This PR makes us not care about that problem.

@ethanbao Feel free to merge upon reviewing; no need to wait on me.


lukesneeringer:
@ethanbao Is this golden test failure unrelated?

ethanbao:
Michael fixed the golden test failure in another PR.

ethanbao:
I started the rebuild of golden test.

BTW, I was not able to see `artifacts` in the circleci report initially. After I click `user settings` and then connect to Google org, which will require me to login again. After the login, I can see the `artifacts` with current org name at top-left, and my github icon at top-right.

lukesneeringer:
> I started the rebuild of golden test.

Getting a different result would depend on merging master first. :-) (Now done.)

PR #281, opened 2017-09-18
Introduce  prefix to allow file path exclusion for src_proto_path field

ethanbao:


geigerj:
@michaelbausor Why do we need such a blacklist in the first place? Can we just change `artman_core.yaml` to whitelist in `src_proto_path` exactly the right subdirectories in `google/api`? How big would the burden be of maintaining such a whitelist? (Presumably it is somewhat rare for new packages to be added to the common protos...)

michaelbausor:
The issue is that `google/api` itself contains protos. So we could perhaps support a whitelist like:
```
src_proto_path:
- "google/api/*.proto"
- "google/api/experimental"
```

Or we could make the whitelist non-recursive, and list all subdirectories, which seems reasonable. I don't have a preference between maintaining a whitelist vs a blacklist (I think maintenance will be roughly equal). What would we do if there was a single proto that we wanted to exclude, rather than a directory? In that case a whitelist approach would be forced to list all protos. This is hypothetical though, so it may not be that strong of an argument.

geigerj:
@michaelbausor Makes sense. This seems fine to me, just wary of single-use-case features.

michaelbausor:
LGTM

Issue #277, opened 2017-09-15
Fix Python style issues, add linting to CI


geigerj:
Running `nox -e lint` shows dozens of Python style violations. The linter environment is not being run in travis, allowing these style violations to accumulate. We should fix the existing violations, then enable the linter in ~travis~ Circle.

ethanbao:
Many lint warnings are currently introduced by main.py, which I am going to drop starting 0.5.*. I will add the linting to CI after that.

lukesneeringer:
I would actually prefer _not_ run pylint, but we should run flake8.

PR #285, opened 2017-09-20
Introduce new semantics of path-typed fields in artman yaml.

ethanbao:
* Any path-type fields specified in configuration YAML parsed by artman may accept relative paths; paths are computed as relative to the file in which the relative path is declared. For example, service_yaml field in artman_pubsub.yaml will become “pubsub.yaml” instead of “google/pubsub/pubsub.yaml” once the artman_pubsub.yaml gets moved to google/pubsub directory.
* The path-typed fields in artman yaml will not support “..” syntax to refer to parent folder.
* --input-dir flag gets renamed to --root-dir, and remain as optional flag (defaults to `.`). Once specified, it will be appended to import_proto_path field of artman config spec at runtime.
* The current import_proto_path field will be removed and no longer configurable from API producer point of view.


artman/cli/main2.py
```diff
@@ -136,13 +136,15 @@ def parse_args(*args):
         help='[Optional] Directory to store output generated by artman. '
         'Default to `./artman-genfiles`', )
     parser.add_argument(
-        '--input-dir',
+        '--root-dir',
         type=str,
-        default='',
+        default=None,
```
lukesneeringer: I got an error previously using a `None` default alongside `type=str`, which is why I made this an empty string. Both are falsy.

artman/config/converter.py
```diff
@@ -170,10 +169,10 @@ def _calculate_git_repos_config(artifact_config, output_dir):
...
 
 def _calculate_proto_paths(proto_paths):
-  src_proto_path, excluded_proto_path = [], []
-  for proto_path in proto_paths:
-      if proto_path.startswith('-'):
-          excluded_proto_path.append(proto_path[1:])
-      else:
-          src_proto_path.append(proto_path)
-  return src_proto_path, excluded_proto_path
+    src_proto_path, excluded_proto_path = [], []
```
lukesneeringer: I do not normally bring up linting issues in PRs, but can we make this be consistent four-space indenting, please?
I will add a linter later.

artman/config/loader.py
```diff
@@ -95,43 +96,38 @@ def _validate_artifact_config(artifact_config):
...
         else:
             normalized_src_proto_paths.append(
-                _normalize_path(src_proto_path, input_dir))
+                _normalize_path(src_proto_path, artman_config_dir))
     artifact_config.src_proto_paths[:] = normalized_src_proto_paths
 
     return artifact_config
 
 def _normalize_path(path, prefix):
+  if '..' in path:
```
lukesneeringer: `if path.startswith('..'):`

lukesneeringer: Also, four-space indenting.

artman/config/loader.py
```diff
@@ -95,43 +96,38 @@ def _validate_artifact_config(artifact_config):
...
+      raise ValueError('Cannot use ".." path-typed field in artman yaml. '
+                       'Please use either path relative to artman yaml '
```
lukesneeringer: This error message very well might be wrong, if it is coming when normalizing a path in a file that is not the artman YAML.

Also, this error should (at minimum) print the actual path that you tried to normalize and raised an exception on. Ideally it would also include the key and the filename, but I recognize that means passing it in and I did not check to see how far down the rabbit hole that would go.

You could also drop the `..` restriction and just roll with it.

artman/config/loader.py
```diff
@@ -95,43 +96,38 @@ def _validate_artifact_config(artifact_config):
...
+                       '(preferred). or an absolute path.')
   return path if os.path.isabs(path) else os.path.join(prefix, path)
```
lukesneeringer: Are we 100% sure that `prefix` was previously passed through `os.path.realpath`?

artman/config/proto/config.proto
```diff
@@ -166,18 +168,14 @@ message Artifact {
...
   // [pubsub_gapic.yaml](https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub_gapic.yaml)
   // as an example. See
   // [config.proto](https://github.com/googleapis/toolkit/blob/master/src/main/proto/com/google/api/codegen/config.proto)
   // for a full GAPIC config spec.
   string gapic_yaml = 10;
 
-  // This is an optional field to specify the root proto_path. It will be
-  // passed to protoc via --proto_path flag. The default value is
-  // "${INPUT_DIR}".
-  repeated string import_proto_path = 11;
```
lukesneeringer:
```proto
reserved 11;
reserved import_proto_path;
```

test/config/testdata/artman_pubsub.yaml
```diff
@@ -3,15 +3,15 @@ common:
   api_version: v1
   organization_name: google-cloud
   src_proto_paths:
-  - google/pubsub/v1
-  - -google/pubsub/v1/excluded
+  - /tmp/input/google/pubsub/v1
+  - -/tmp/input/google/pubsub/v1/excluded
```
lukesneeringer: What is the deal with this? Also, `- -` syntax in YAML scares me. Is that a string with a minus at the beginning, or a nested list? Will different YAML parsers even all do the same thing?

artman/config/converter.py
ethanbao: I guess you made the comment before I pushed another commit. The linter did warn me, though sometime it is hard to track given we have already had many linter warnings.

artman/cli/main2.py
ethanbao: My bad. Should have changed this.

artman/config/loader.py
ethanbao: Done. Failed to catch this in the linter report.

artman/config/proto/config.proto
ethanbao: Good catch.

test/config/testdata/artman_pubsub.yaml
ethanbao: It is a string with a minus at the beginning. It is a syntax to exclude certain file or folder.

artman/config/loader.py
ethanbao: Keep the ".." restriction for now. I am fine to lift it if it turns out to be a bad restriction.

PTAL

ethanbao:
Fixed the unittest failure. It turned out to be a transitive dependency (networkx) launched a new version (2.0) today which caused the failure. Forced the version to be 1.11.

artman/cli/main2.py
```diff
@@ -136,13 +136,15 @@ def parse_args(*args):
...
         type=str,
         default='',
         help='[Optional] Directory with all input that is needed by artman, '
-             'which include but not limited to API protos, service config '
-             'yaml, and GAPIC config yaml. Default to the settings from user '
-             'configuration.',
+        'which include but not limited to API protos, service config yaml '
+        'and GAPIC config yaml. It will be passed to protobuf compiler via '
+        '`-I` flag in order to generate descriptor. If not specified, it '
+        'will first look up in artman user config. If not found, an error '
```
geigerj: "will first look up in the artman config" -> "will derive the root directory from the source proto files"

artman/cli/main2.py
```diff
@@ -136,13 +136,15 @@ def parse_args(*args):
...
     parser.add_argument(
-        '--input-dir',
+        '--root-dir',
         type=str,
         default='',
         help='[Optional] Directory with all input that is needed by artman, '
-             'which include but not limited to API protos, service config '
-             'yaml, and GAPIC config yaml. Default to the settings from user '
-             'configuration.',
+        'which include but not limited to API protos, service config yaml '
```
geigerj: nit: "include but not limited to" -> "includes but is not limited to"

artman/config/converter.py
geigerj: I filed an issue for adding a linter to CI last week. There are dozens of already existing violations. https://github.com/googleapis/artman/issues/277

artman/cli/main2.py
```diff
@@ -54,7 +54,7 @@ def main(*args):
     # Get to a normalized set of arguments.
     flags = parse_args(*args)
     user_config = read_user_config(flags)
-    _adjust_input_dir(flags.input_dir)
+    _adjust_root_dir(flags.root_dir)
```
geigerj: Is this a no-op `if not flags.root_dir`?

artman/config/loader.py
```diff
@@ -95,43 +96,48 @@ def _validate_artifact_config(artifact_config):
...
     return artifact_config
 
-def _normalize_path(path, prefix):
-  return path if os.path.isabs(path) else os.path.join(prefix, path)
+
+def _normalize_path(path, artman_config_path, field):
+    if path.startswith('..'):
+        raise ValueError(
+            '".." is disallowed in `%s` field of `%s`. Please use either a '
+            'path relative to `%s` (preferred). or an absolute path.'
```
geigerj: nit: extra period, `(preferred).` -> `(preferred)`

artman/config/loader.py
```diff
@@ -95,43 +96,48 @@ def _validate_artifact_config(artifact_config):
...
+                                'src_proto_paths'))
     artifact_config.src_proto_paths[:] = normalized_src_proto_paths
 
     return artifact_config
 
-def _normalize_path(path, prefix):
-  return path if os.path.isabs(path) else os.path.join(prefix, path)
+
+def _normalize_path(path, artman_config_path, field):
+    if path.startswith('..'):
```
geigerj: `or path.contains('/..')`? Otherwise, you can get around this just by `./../foo`

artman/cli/main2.py
ethanbao: Updated.

artman/cli/main2.py
ethanbao: I didn't implement that feature here. Unlike google apis protos, there is no guarantee it will be correct.

artman/config/converter.py
ethanbao: Will address that issue in a follow-up.

artman/cli/main2.py
ethanbao: There should always have one after normalization. Updated line 290 to normalize this flag.

artman/config/loader.py
ethanbao: Updated.

artman/config/loader.py
ethanbao: Good catch.

artman/cli/main2.py
lukesneeringer: I thought so. Checking again.

lukesneeringer: Confirmed what @ethanbao just said.

artman/cli/main2.py
geigerj: I am not sure I understood your plan. Do you plan to put this logic in g3artman?

I think it makes sense to implement that feature in artman, not g3artman. It then becomes an artman convention to structure your proto source directories according to your proto packages.

artman/cli/main2.py
```diff
@@ -279,11 +281,14 @@ def normalize_flags(flags, user_config):
...
 
-    # Save the input directory back to flags if it was not explicitly set.
-    if not flags.input_dir:
-        flags.input_dir = pipeline_args['local_paths']['googleapis']
+    if flags.root_dir:
+        root_dir = flags.root_dir
+    elif pipeline_args['local_paths']['googleapis']:
+        root_dir = pipeline_args['local_paths']['googleapis']
+    else:
+        root_dir = os.getcwd()
```
geigerj: I thought you and @lukesneeringer agreed in offline discussion not to default to the working directory?

artman/config/loader.py
```diff
@@ -95,43 +96,48 @@ def _validate_artifact_config(artifact_config):
...
     return artifact_config
 
-def _normalize_path(path, prefix):
-  return path if os.path.isabs(path) else os.path.join(prefix, path)
+
+def _normalize_path(path, artman_config_path, field):
+    if path.startswith('..') or '/..' in path:
+        raise ValueError(
+            '".." is disallowed in `%s` field of `%s`. Please use either a '
+            'path relative to `%s`, (preferred). or an absolute path.'
```
geigerj: There is still an extra period here?

artman/cli/main2.py
ethanbao: Yes, we did. This was an accidental revert of my previous change because I switched my working environment in between.

artman/cli/main2.py
geigerj: Per offline discussion, considering this resolved, since using such logic relies on potentially unstable assumptions about the directory structure of your source proto path.

PR #289, opened 2017-09-26
Fix protos in subdirectories for Python.

lukesneeringer:
This is a theoretically-correct fix for the "protos in subdirectories" problem that Jacob and I discovered on Firestore today.

Note: The GAPIC does not generate the correct path for subdirectory cases (this will be a toolkit bug for @landrito).

Note also: There is at least one semantic change, which is that proto path transformation will not happen cross-API anymore. This is probably okay (we do not rely on this I doubt there was any world where it ever worked) but needs some discussion with @geigerj and @landrito.

### DIFF
--- a/artman/tasks/python_grpc_tasks.py
+++ b/artman/tasks/python_grpc_tasks.py
@@ -55,8 +55,8 @@ class PythonChangePackageTask(task_base.TaskBase):
 
     # TODO (geigerj): add regex for documentation link updates?
 
-    def execute(self, src_proto_path, import_proto_path, common_protos_yaml,
-                organization_name):
+    def execute(self, api_name, api_version, src_proto_path, import_proto_path,
+                common_protos_yaml, organization_name):
         self._organization_name = organization_name
 
         with io.open(common_protos_yaml) as file_:
@@ -78,9 +78,13 @@ class PythonChangePackageTask(task_base.TaskBase):
         new_import_path = [new_proto_dir]
 
         self._copy_and_transform_directories(
-            src_proto_path, new_proto_dir, common_protos, paths=new_src_path)
+            src_proto_path, new_proto_dir, common_protos, paths=new_src_path,
+            api_name=api_name, api_version=api_version,
+        )
         self._copy_and_transform_directories(
-            import_proto_path, new_proto_dir, common_protos)
+            import_proto_path, new_proto_dir, common_protos,
+            api_name=api_name, api_version=api_version,
+        )
 
         # Update src_proto_path, import_proto_path
         return list(new_src_path), new_import_path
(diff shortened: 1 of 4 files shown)


artman/tasks/python_grpc_tasks.py
```diff
@@ -96,45 +100,49 @@ def _extract_base_dirs(self, proto_file):
...
         """Transform to the appropriate proto package layout.
 
         Works with arbitrary separator (e.g., '/' for import statements,
         '.' for proto package statements, os.path.sep for filenames)
         """
+        # Split the proto path into a list, and move the `.proto` extension
+        # off to the side for now.
         if sep != '.' and pkg.endswith('.proto'):
-            dotted = pkg[:-6].replace(sep, '.')
+            dotted = pkg[:-6].replace(sep, '.').split('.')
```
geigerj: nit: this variable name doesn't make sense any more. Also, it is easier to read as
`paths = pkg[:-len('.proto')].split(sep)`

artman/tasks/python_grpc_tasks.py
```diff
@@ -96,45 +100,49 @@ def _extract_base_dirs(self, proto_file):
...
+        # this is a no-op.
+        if api_name not in dotted or api_version not in dotted:
+            return pkg
 
         # Special case: If the organization name is "google-cloud", then we
         # have to ensure that "cloud" exists in the path. The protos
         # themselves may not follow this.
-        if 'cloud' not in dotted and self._organization_name == 'google-cloud':
-            dotted = dotted.replace('google.', 'google.cloud.', 1)
+        if all(['google' in dotted,
```
geigerj: nit: just use `and`?

artman/tasks/python_grpc_tasks.py
```diff
@@ -96,45 +100,49 @@ def _extract_base_dirs(self, proto_file):
...
         if sep != '.' and pkg.endswith('.proto'):
-            dotted = pkg[:-6].replace(sep, '.')
+            dotted = pkg[:-6].replace(sep, '.').split('.')
             suffix = '.proto'
         else:
-            dotted = pkg.replace(sep, '.')
+            dotted = pkg.replace(sep, '.').split('.')
             suffix = ''
 
-        # Sanity check: Do not transform common protos.
```
geigerj: Does this mean we don't need to special case common protos any more?

artman/tasks/python_grpc_tasks.py
```diff
@@ -96,45 +100,49 @@ def _extract_base_dirs(self, proto_file):
...
-        for common_pkg in common_protos:
```
geigerj: Also, do we really want to no-op for unversioned, *non*-common protos (e.g., `google.appengine.legacy`; I do not know if there are any examples in actual current codegen use), or do we still want to do the "cloud" transformation below?

artman/tasks/python_grpc_tasks.py
```diff
@@ -143,29 +151,37 @@ def _copy_proto(self, src, dest, common_protos):
...
                 if paths is not None:
                     paths.add(sub_new_src)
 
                 dest = os.path.join(sub_new_src, os.path.basename(proto))
                 if not os.path.exists(dest):
                     self.exec_command(['mkdir', '-p', sub_new_src])
                 self._copy_proto(
-                    proto, os.path.join(sub_new_src, dest), common_protos)
+                    proto, os.path.join(sub_new_src, dest), common_protos,
+                    api_name=api_name, api_version=api_version,
```
geigerj: nit: inconsistent to use kwargs only for this change?

artman/tasks/python_grpc_tasks.py
lukesneeringer: This is part of why I marked it `do not merge`; we need to talk about this aspect. The current logic will bypass any case where `api_name` and `api_version` are not present in the proto path. This does mean that common protos are no longer a special case.

However, there is another effect here, which is that transitive APIs would no longer get caught either. We had test cases for `google.service.v1` and `google.other_api.v2`, and it is no longer the case that the latter of the two would be transformed at all.

We need to discuss whether this is a case we actually needed, because if so, I just created a problem. :-)

artman/tasks/python_grpc_tasks.py
lukesneeringer: Correct; I was in a hurry and not going back and figuring out what everything else on these private methods were.

lukesneeringer:
Note: This PR _is still active_; I will get it to working after `${tomorrow}` is done.

ethanbao:
I don't know python library output structure enough to judge the transformation. The artman code looks good, though the smoketest is failing (I bet it is a small update of library-example.golden file).

lukesneeringer:
I figured out what was wrong with the golden test, but it exposed an edge case that is wrong in my code too (it would break video intelligence right now).

ethanbao:
Close this PR for now.

PR #293, opened 2017-09-29
C# Move csproj files to the correct location

chrisdunelm:


artman/tasks/gapic_tasks.py
```diff
@@ -145,8 +145,9 @@ def execute(self, gapic_code_dir, grpc_code_dir, proto_code_dir, gapic_api_yaml)
         self.exec_command(['sh', '-c', 'cp {0}/*.cs {1}'.format(grpc_code_dir, prod_dir)])
         self.exec_command(['sh', '-c', 'cp {0}/*.cs {1}'.format(proto_code_dir, prod_dir)])
         self.exec_command(['sh', '-c', 'mv {0}/*.cs {1}'.format(gapic_code_dir, prod_dir)])
+        self.exec_command(['sh', '-c', 'mv {0}/*.csproj {1}'.format(gapic_code_dir, prod_dir)])
         self.exec_command(['sh', '-c', 'mv {0}/*Snippets.g.cs {1}'.format(prod_dir, snippets_dir)])
-        
+
```
jskeet: Just remove the line entirely, maybe?

chrisdunelm: I think I remember that having two blank lines between classes is normal...
(But I could be wrong)

jskeet: Fair enough - certainly not a problem.

PR #295, opened 2017-09-29
Update smoketest_artman.py to use the new artman config locations.

ethanbao:


ethanbao:
Ignore the golden test failure which will be fixed in another PR.

test/smoketest_artman.py
```diff
@@ -16,73 +16,75 @@
...
-    if apis:
-        for api in apis.split(','):
-            artman_config_whitelist.append('artman_%s.yaml' % api)
+def run_smoke_test(root_dir):
+    failure = []
+    success = []
+    for artman_yaml_path in glob.glob('%s/google/**/artman_*.yaml' % root_dir,
+                                      recursive=True):
+        if os.path.relpath(artman_yaml_path, root_dir) in CONFIG_BLACKLIST:
+            print('Skip %s' % artman_yaml_path)
```
geigerj: I am suspicious of using `print` in a test, especially since many test frameworks swallow stdout. Where will these `print` messages be captured?

ethanbao: Though it is called smoketest, I don't plan to use any test framework. But I would love to follow the standard practice. Shall I use logging then?

ethanbao:
Gently ping

test/smoketest_artman.py
```diff
@@ -16,98 +16,117 @@
...
-    grpc_pipeline_args = [
-        'artman2',
-        '--local',
-        '--config', artman_config,
-        '--input', input_dir,
-        'generate', artifact_name
-    ]
-    return subprocess.call(grpc_pipeline_args, stdout=subprocess.PIPE)
+def _generate_gapic_library(artman_config, artifact_name, root_dir, log_file):
+    with open(log_file, "a") as log:
```
geigerj: nit: single quotes (`'a'`)

test/smoketest_artman.py
```diff
@@ -16,98 +16,117 @@
...
 
+def _setup_logger(log_file):
+    """Setup logger with a logging FileHandler."""
+    log_file_handler = logging.FileHandler(log_file)
+    logger.addHandler(log_file_handler)
+    logger.addHandler(logging.StreamHandler())
+    return log_file
+
+
 if __name__ == "__main__":
```
geigerj: nit: single quotes

test/smoketest_artman.py
```diff
@@ -16,98 +16,117 @@
...
-    input_dir = os.path.abspath(input_dir)
-    artman_config_dirs = [
-        os.path.join(input_dir, 'gapic', 'api'),
-    ]
-
-    artman_config_whitelist = []
-    if apis:
-        for api in apis.split(','):
-            artman_config_whitelist.append('artman_%s.yaml' % api)
+CONFIG_BLACKLIST = [
```
geigerj: I'm a little concerned about the maintainability of such a blacklist. Will we remember to update once these should succeed? It would be good if each line linked to an issue describing what is failing and what needs to be done to make it succeed.

test/smoketest_artman.py
```diff
@@ -16,98 +16,117 @@
...
+                                       root_dir,
+                                       log_file):
+                msg = 'Failed to generate %s of %s.' % (
+                    artifact.name, artman_yaml_path)
+                failure.append(msg)
+            else:
+                msg = 'Succeded to generate %s of %s.' % (
+                    artifact.name, artman_yaml_path)
+                success.append(msg)
+            logger.info(msg)
```
geigerj: It seems like this double-logs everything? (line 73 + 75/77)

ethanbao: It is intended. Because the test takes too long to complete and I usually want to get an update in the mid, I found it is helpful to get a peek before the test is fully complete. Actually, circleci will by default kill the task after 10mins if nothing prints out. While that is time threshold configurable, I find it is easier to just print it out at each iteration, and if one generation task does hold longer than 10 mins, it probably means there is something wrong, and I want the task to stop after 10 mins instead of waiting even longer.

test/smoketest_artman.py
ethanbao: Done.

test/smoketest_artman.py
ethanbao: I share your concern. I don't plan to file GitHub issues as there won't be assignees except myself. I do have a plan to address each of them. Another option is to run generation for those in the blacklist but not count them as errors but warnings if the execution fails.

test/smoketest_artman.py
geigerj: In that case, let's just `logger.info` or `logger.error` inline; no need to repeat them at the end.

test/smoketest_artman.py
geigerj: Makes sense. Could you either:

- file an uber-issue to remove this blacklist once each case is addressed
- run these generations, and only warn, not fail

so that we'll have some tracking?

test/smoketest_artman.py
ethanbao: The part in the end is a summary. The smoketest writes everything to a log that will be viewable in circleci artifact section. That part includes both messaged logged by logger plus artman execution output (see the subprocess.call part). However, only the logger part is printed out to console/streamhandler. A full report without summary in the end is almost unreadable in this case.

geigerj: Ack. Did not fully understand what the log will look like. That seems reasonable.

ethanbao: A report in console: https://circleci.com/gh/ethanbao/artman/171?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

A full report in artifacts: https://171-57229230-gh.circle-artifacts.com/0/tmp/reports/smoketest.log

ethanbao:
Added warning section.

PR #296, opened 2017-09-29
Bump artman verison 0.4.20 to contain the c# csproj generation support

ethanbao:


ethanbao:
Need to release a new version to capture Chris's https://github.com/googleapis/artman/pull/293

PR #298, opened 2017-10-06
Introduce a "build_and_push_docker_image" job in artman/circleci

ethanbao:
This is a example when the tasks runs:
https://circleci.com/gh/ethanbao/artman/410

The smoketest (all) will fail, but I manually disabled the smoketest and confirmed that the deploy step will work (see https://hub.docker.com/r/googleapis/artman/tags/ for an auto-build artman docker image).


.circleci/config.yml
```diff
@@ -2,13 +2,73 @@
...
+      - save_cache:
+          key: artman-{{ .Branch }}
+          paths:
+            - /caches/artman.
+      - run:
+          name: Run golden tests
+          command: |
+            docker run -it -e RUNNING_IN_ARTMAN_DOCKER=True -v /tmp/reports:/tmp/reports artman /bin/bash -c "pip install pytest; git clone https://github.com/googleapis/googleapis /artman/test/golden/googleapis; py.test /artman/test/golden/artman_golden_test.py --googleapis-dir=/artman/test/golden/googleapis"
+      # When build performance is a concern, consider separting this job into
+      # a workflow with multiple jobs and have the smoketest run in parrall
```
geigerj: "parrall" -> "parallel"

.circleci/config.yml
```diff
@@ -2,13 +2,73 @@
...
+            docker save -o /caches/artman.tar artman
+      - save_cache:
+          key: artman-{{ .Branch }}
+          paths:
+            - /caches/artman.
+      - run:
+          name: Run golden tests
+          command: |
+            docker run -it -e RUNNING_IN_ARTMAN_DOCKER=True -v /tmp/reports:/tmp/reports artman /bin/bash -c "pip install pytest; git clone https://github.com/googleapis/googleapis /artman/test/golden/googleapis; py.test /artman/test/golden/artman_golden_test.py --googleapis-dir=/artman/test/golden/googleapis"
+      # When build performance is a concern, consider separting this job into
```
geigerj: "separting" -> "separating"

.circleci/config.yml
```diff
@@ -2,13 +2,73 @@
...
+      # a workflow with multiple jobs and have the smoketest run in parrall
+      # with additional paied containers.
```
geigerj: "paied" -> "paid"

.circleci/config.yml
```diff
@@ -2,13 +2,73 @@
...
+      - run:
+          name: Run smoketests
+          command: |
+            mkdir /tmp/reports
+            echo "this is a test" >> /tmp/reports/smoketest.log
```
geigerj: > `>> /tmp/reports/smoketest.log`

Is it expected that smoketest.log already exists? If so, we should be more specific, for example, "Artman smoketest at {commit_id}, {time}". If not, no need to append to the log file.

.circleci/config.yml
```diff
@@ -10,6 +10,75 @@ jobs:
...
 
+  build_and_push_docker_image:
+    working_directory: /usr/src/artman
+    docker:
+      - image: docker:17.05.0-ce-git
+    steps:
+      - checkout
+      - setup_remote_docker
+      - run:
+          name: Finger crossed!
```
geigerj: nit: "finger" -> "fingers" :)

ethanbao: Woops, thanks for making it a bit easier.

.circleci/config.yml
ethanbao: Done

.circleci/config.yml
ethanbao: Done

.circleci/config.yml
ethanbao: Done

.circleci/config.yml
ethanbao: Removed this line. I prefer to use constant file name for two reasons: 1) each execution will uses a new container, so it is not going to cause conflicts. 2) circleci 2.0 currently has a limited test analysis/reporting system, and I plan to create a GAE-based job which will trigger cadence test via cron task, and collect test result by checking circleci artifacts. Using fixed file name will make the artifacts fetching a bit easy.

.circleci/config.yml
```diff
@@ -10,6 +10,77 @@ jobs:
...
+      only:
+        - master
+    steps:
+      - checkout
+      - setup_remote_docker
+      - run:
+          name: Fingers crossed!
+          command: |
+            echo "Fingers crossed!"
+      - restore_cache:
```
lukesneeringer: Unless you get a substantial speed improvement, I would probably omit this (and `save_cache`).

ethanbao:
The performance gain is substantial. Saved 20+ mins in building time. More importantly, it is the same cached layer when pushing to Dockerhub, meaning the downstream users will not need to download the whole image for a new docker image in most cases.

ethanbao:
Merged, thank you for the review. There is a similar CI PR in googleapis repo: https://github.com/googleapis/googleapis/pull/474. PTAL.

While running the same smoketest script, the googleapis PR uses different artman docker image for different purpose.

PR #300, opened 2017-10-10
Bump artman version to 0.5.0

ethanbao:
Changes include:
* Drop legacy artman main.py
* Renamed main2.py to main.py
* Removed part of golden tests that tests legacy artman yamls.
* Renamed test/golden/artman_library_example_new.yaml to artman_library_example.yaml
* Update smoketest to use `artman` instead of `artman2`.
* Update unit tests to test the new artman CLI syntax.


artman/cli/main.py
```diff
@@ -11,34 +11,41 @@
...
 
+from artman.config import converter, loader
+from artman.config.proto.config_pb2 import Artifact, Config
 from artman.cli import support
 from artman.pipelines import pipeline_factory
-from artman.utils import job_util, pipeline_util, config_util
+from artman.utils import config_util
 from artman.utils.logger import logger, setup_logging
 
+ARTMAN_DOCKER_IMAGE = 'googleapis/artman:0.5.0'
```
geigerj: (doesn't have to be addressed in this PR): will this ever be if a format other than `googleapis/artman:{latest-version}`?

if not, consider making the version a global constant and referencing it both from here and `setup.py` to avoid having to update it in two places.

test/golden/library_example.golden
```diff
@@ -273,10 +273,10 @@
 /js/library_example-v1/package.json
 /js/library_example-v1/protos/google/example/library/v1/library.proto
 /js/library_example-v1/src/index.js
-/js/library_example-v1/src/v1/doc/doc_library.js
 /js/library_example-v1/src/v1/index.js
 /js/library_example-v1/src/v1/library_service_client.js
 /js/library_example-v1/src/v1/library_service_client_config.json
+/js/library_example-v1/src/v1/doc/google/example/library/v1/doc_library.js
```
ethanbao: @lukesneeringer, @landrito Can you confirm this is expected change?

artman/cli/main.py
ethanbao: I don't have a plan to change the format. It makes sense to make version a constant. I didn't realize that setup.py is a python file itself.

PR #301, opened 2017-10-10
Update PHP formatters and format calls

saicheems:
- Remove the second `php-cs-fixer` call for "phpdoc_var_to_type", which
  is now obsolete.


Dockerfile
```diff
@@ -199,4 +199,4 @@ RUN mkdir -p /root/
 ADD artman-user-config-in-docker.yaml /root/.artman/config.yaml
 
 # Install artman.
-RUN pip3 install googleapis-artman==0.4.20
+RUN pip3 install googleapis-artman==0.5.0
```
ethanbao: Let's not bump the minor version to 5.

saicheems: Done

PR #306, opened 2017-10-17
Keep an empty WHITEST in artman smoketest

ethanbao:
The artman yamls in the whitelist were either fixed or removed.


ethanbao:
Let me keep the whitelist for a while and can do a final cleanup if we don't see a valid case where an artman yaml should be whitelisted.

ethanbao:
Rerun the unit test fixed the problem. Might be some circleci flakyness

PR #313, opened 2017-10-31
Update docker image

geigerj:
Previous image incompatible with `config_schema_version` in GAPIC yaml

### DIFF
--- a/artman/cli/main.py
+++ b/artman/cli/main.py
@@ -40,7 +40,7 @@ from artman.pipelines import pipeline_factory
 from artman.utils import config_util
 from artman.utils.logger import logger, setup_logging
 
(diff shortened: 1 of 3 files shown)


andreamlin:
There is a newer docker image googleapis/artman:0.5.0-rc03. Why isn't that one hardcoded in?

geigerj:
@andreamlin I think it was just that @ethanbao hadn't released that image yet at the time I created this PR. @ethanbao Can we update this hardcoded value as part of the docker image release process?

ethanbao:
It is possible. We actually have a release doc, I need to figure out a way to enforce that.

ethanbao:
OTOH, any problem with using 0.5.0-rc02 for now, @andreamlin? There is no guarantee that the latest pushed docker image will be marked as the one used in artman.

PR #323, opened 2017-11-03
Fix a bug in artman that might not pass the root dir from host to Artman Docker container.

ethanbao:
Will introduce a smoketest to capture the artman execution running in docker mode. Current artman CI tests are already running inside the Artman Docker container, which is why we didn't capture this until customer reports the issue.


andreamlin:
:D

PR #327, opened 2017-11-07
Fix couple of issues related to artman publish

ethanbao:
* support.py will no longer set default value to "api_client_staging" in user config. Instead, it will clone the repo to output dir if such config is not available.
* Several cleanup will only be executed if the execution is done in local mode. Otherwise, it will run into permission error when deleting mounted folder.
* Fix unit test
* Switch to use `artman publish --dry-run` in CI smoketests, which includes generation step.
* Add CI test to cover artman execution in Docker mode.


artman/tasks/publish/local.py
```diff
@@ -53,12 +53,19 @@ def execute(self, git_repo, local_paths, output_dir,
...
         repo_name_underscore = repo_name.replace('-', '_')
-        api_repo = local_paths.get(repo_name_underscore,
-            os.path.join(local_paths.get('reporoot', '..'), repo_name),
-        )
-        api_repo = os.path.realpath(os.path.expanduser(api_repo))
+        api_repo = local_paths.get(repo_name_underscore)
+        if not api_repo:
+            api_repo = os.path.join(output_dir, repo_name)
+            if os.path.exists(api_repo):
+                self.exec_command(['rm', '-rf', api_repo])
```
michaelbausor: Not completely sure, but does this cause a danger that someone's existing repo clone (that might contain local branches/feature work that hasn't been pushed) will get deleted and replaced with a clean checkout?

ethanbao: It could in theory. We will wipe out the content only if such local setting is not specified in artman user config, and the final api_repo would be under output_dir folder. Usually, we don't expect user to mix the input and output directory.

ethanbao: For example, the default output location would be ./artman-genfiles, and therefore a "api-client-staging" repo will be under ./artman-genfiles/api-client-staging.

michaelbausor: Ah ok, that makes sense

artman/cli/main.py
```diff
@@ -439,7 +439,6 @@ def _run_artman_in_docker(flags):
     root_dir = flags.root_dir
     output_dir = flags.output_dir
     artman_config_dirname = os.path.dirname(flags.config)
-    user_config = os.path.join(os.path.expanduser('~'), '.artman')
```
geigerj: I assume this is just because none of the user config is needed when running in docker?

artman/config/converter.py
```diff
@@ -172,7 +172,11 @@ def _calculate_git_repos_config(artifact_config, output_dir):
                 path['dest'] = map_entry.dest
             if map_entry.name:
                 path['artifact'] = map_entry.name
-            paths.append(path)
+            # The "proto" and "grpc" publishing artifact type is only used in
+            # Java. Ignore such configuration if defined for other languages.
```
geigerj: Shouldn't this raise a warning? Silently ignoring the paths could be confusing.

artman/tasks/publish/local.py
geigerj: > Usually, we don't expect user to mix the input and output directory.

To be clear, this is an issue only if the input is under an `artman-genfiles` directory, right?

test/tasks/publish/test_local.py
```diff
@@ -116,32 +116,77 @@ def test_execute_with_grpc(self, is_dir, success, exec_command):
...
+                'reporoot': '/rr',
+                'api_client_staging': '/rr/api-client-staging'
             },
-            local_paths={'reporoot': '/rr'},
             output_dir='/tmp/out',
         )
 
         # Establish the commands we expect to have been called.
-        expected_commands = (
+        expected_commands = [
```
geigerj: Why change immutable to mutable? Usually we prefer immutable if possible.

artman/tasks/publish/local.py
ethanbao: No. If one specify a output_dir explicitly, and that output_dir contains input files, it might still be affected.

geigerj:
@cbao Looks good to me other than some uncertainty about the `rm -rf api_repo`. What is the situation in which this could delete the user's input?

ethanbao:
When you run `artman publish` without specifying the local repo location through artman user config. It will clone the dest repo to {output_dir}/{repo_name}. If the directory already exists, it will be wiped out. That way, you can repeat the execution of `artman publish` during release or development, without manually cleanup the folder.

geigerj:
@ethanbao That actually does seem dangerous. It would be rare that that ever happens, but it would be really bad if we wiped out someone's local git repository, including potentially local branches, etc. I'd prefer we just raised an error if the dir already exists.

PR #332, opened 2017-11-10
Do not hard-code google in the namespace path move task.

lukesneeringer:
There is no guarantee that a namesapce starts with google.


lukesneeringer:
@ethanbao This _should_ be safe but I want to see what the tests do. (Needed to correctly serve things that are non-cloud in Python.)

lukesneeringer:
Update: They are safe.

PR #335, opened 2017-11-14
Revert "Do not hard-code google in the namespace path move task."

ethanbao:
Reverts googleapis/artman#332


ethanbao:
The PR is not the culprit, close this PR.

PR #340, opened 2017-11-21
WIP: Bump grpc plugin version

michaelbausor:


ethanbao:
Is it intended to bump toolkit and googleapis at the same time?

michaelbausor:
I was simply bumping them to the latest versions, in order to include all of the most recent changes. Is that what we normally do? I am happy to set them to some other value.

The change to the php grpc plugin should not affect any other language, but the changes to toolkit and googleapis clearly could, depending on what changes those languages have in toolkit.

ethanbao:
Let's revert the two changes from this PR so that this PR is less likely to cause problem and be rollbacked.

I will bump artman to 0.5.4, along with which I will bump the toolkit and googleapis versions.

michaelbausor:
OK sounds good

michaelbausor:
PTAL

PR #347, opened 2017-12-04
Fix golden test failure

ethanbao:
This way, one can easily diff the golden test result and have the test failure fixed.

Besides, Include actual golden test report in circleci artifacts.


ethanbao:
Merging this PR as the golden test change is expected.

PR #349, opened 2017-12-15
Rev artman to 0.5.5 to get new toolkit, googleapis.

lukesneeringer:


ethanbao:
Change this file too: https://github.com/googleapis/artman/blob/master/artman/cli/main.py#L44

Golden test is basically broken, circle CI doesn't download the latest artman image (probably due to cache). Update library_example.golden to whatever is here: https://4684-55101236-gh.circle-artifacts.com/0/usr/src/artman/test/golden/actual_library_example.golden

lukesneeringer:
> Golden test is basically broken, circle CI doesn't download the latest artman image (probably due to cache). Update library_example.golden to whatever is here: https://4684-55101236-gh.circle-artifacts.com/0/usr/src/artman/test/golden/actual_library_example.golden

As in, the test no longer serves its purpose? Then we should de-require it rather than update to a bad value.

PR #355, opened 2017-12-21
Artman documentation

lukesneeringer:
Rewritten from scratch.


README.rst
```diff
@@ -1,140 +1,47 @@
 Google API Artifact Manager
 ===========================
 
-Google API Artifact manager (artman) is a set of modules used to automate the
-creation of software artifacts related to APIS defined using `protobuf`_ IDL.
+Google API Artifact Manager (``artman``) is a program used to automate the
+creation of API client libraries.
```
ethanbao: generation and publishing (to align with `generate` and `publish` commands we have?)

README.rst
```diff
@@ -1,140 +1,47 @@
...
 
-artman is an extensible framework that is responsible for creating all artifacts
-related to an API including
+In order to be consumed by artman, APIs require:
 
-- distribution packages in all supported programming languages
-- generic documentation websites
-- language-specific documentation websites (javadoc, readthedocs, etc)
+  * A `Protocol Buffers`_ description of the API.
```
ethanbao: nit: Shall we say we only support proto3?

README.rst
```diff
@@ -1,140 +1,47 @@
...
+  * A `service configuration`_ stub. This is YAML configuration which
+    designates how the API is housed within Google's infrastructure.
```
ethanbao: Would mentioning of "Google" make reader feel that artman is google-only? I understand this is the status quo. But we might want to lift that limitation in 2018. Your call.

One way I personally differentiate the service config and gapic config is that one is server side config and the other is client side. So maybe say "how the API server side is configured within Google's infrastructure".

README.rst
```diff
@@ -1,140 +1,47 @@
...
-License
--------
+Documentation
+=============
 
-BSD - See `LICENSE`_ for more information.
+Documentation is available on `Read the Docs`_.
 
-.. _`LICENSE`: https://github.com/googleapis/artman/blob/master/LICENSE
+.. _`Read the Docs`: https://googleapis-artman.readthedocs.io/
```
ethanbao: The link is broken. I assume it will soon become available once the PR is merged and uploaded to readthedocs?

docs/installing.rst
```diff
@@ -0,0 +1,84 @@
...
+.. _`Dockerfile`: https://github.com/googleapis/artman/blob/master/Dockerfile
+
+googleapis
+~~~~~~~~~~
+Currently, artman is primarily used for building Google Cloud client libraries,
+and depends on a large, interdependent configuration structure. This is housed
+in the `googleapis`_ repository.
+
+In order to run artman, you will need to clone this repository, and generally
+you will want to be in this directory when invoking artman.
```
ethanbao: s/will want to/should.

This is the current behavior though it is not by design. I need to fix an issue in artman.

docs/support.rst
```diff
@@ -0,0 +1,39 @@
+Support
+=======
+
+Python Versions
+---------------
+
+artman is currently tested with Python 2.7, Python 3.4, Python 3.5,
```
ethanbao: A question not related to this PR. What if we say artman doesn't support python 2.7?

docs/usage.rst
```diff
@@ -0,0 +1,47 @@
...
+  * ``python_gapic`` is the objective -- the thing you want to be produced.
+    You essentially always want a GAPIC (an old code name for the complete,
+    auto-generated API client), but seven languages are supported at this
+    time; ``ruby_gapic``, ``java_gapic``, etc. will give you what you expect.
+
+The package will be dropped in a newly created ``artman-genfiles/`` directory
+within your current working directory; the output of the command will tell
+you precisely where it put the library.
+
+.. _`Natural Language API`: https://cloud.google.com/natural-language/
```
ethanbao: FYI. I will later add a reference section with details for each individual subcommand and flags.

docs/usage.rst
```diff
@@ -0,0 +1,47 @@
...
+        consistently having them.)
+
+  * ``generate`` is the command / verb. (artman also supports ``publish``,
+    which is able to, for example, automatically create a pull request on
+    GitHub.)
+    
+  * ``python_gapic`` is the objective -- the thing you want to be produced.
+    You essentially always want a GAPIC (an old code name for the complete,
+    auto-generated API client), but seven languages are supported at this
+    time; ``ruby_gapic``, ``java_gapic``, etc. will give you what you expect.
```
ethanbao: I suggest we mention that `python_gapic` is the naming convention, and one can define its artifact to any name as long as it is unique inside one single artman yaml.

README.rst
lukesneeringer: > Would mentioning of "Google" make reader feel that artman is google-only? I understand this is the status quo. But we might want to lift that limitation in 2018. Your call.

I want to honestly document the current state of the world. This fact is obviously something I want to change, but we do nobody any favors by obscuring how it works now.

README.rst
lukesneeringer: Correct. I already set up RTD but it builds from master.

docs/support.rst
lukesneeringer: Sadly, Python 2.7 is still system Python for Goobuntu and gLinux, as well as many other distributions. Dropping support for 2.7 would increase the barrier to usage pretty dramatically.

docs/usage.rst
lukesneeringer: This is intended to be a getting started doc -- allow a user with no previous knowledge of the system build an already-defined API. All APIs right now are defined by this convention.

docs/usage.rst
lukesneeringer: Agreed. We also need a section defining the configuration. But, one step at a time. :-)

README.rst
lukesneeringer: What happens if you send proto2? Does API Compiler not support it?

README.rst
lukesneeringer: The docs as built from this branch are [here](http://googleapis-artman.readthedocs.io/en/readme/).

docs/support.rst
ethanbao: Ack. Makes sense.

README.rst
ethanbao: I will get you to an uncharted area. No one knows what could possibly go wrong when you mix proto2 and proto3 in GRPC world.

README.rst
ethanbao: Fair. Given it is under googleapis org anyway.

docs/usage.rst
ethanbao: I am fine, though this doc is usage.rst not getting-started.rst.

PR #364, opened 2018-01-05
 Introduce optional --local-repo-dir subflag to `publish --dry-run` command

ethanbao:
Introduce optional --local-repo-dir subflag to `publish --dry-run` command, which points to a local github repo directory, that will then be mounted to docker container. Then google-cloud-java team can easily write a 20-line python script (excluding constant string list pointing to different artman yaml) to do batch refresh. And any customization, if necessary, can be done in the repo-specific batch script without coupling with artman.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -189,7 +189,9 @@ jobs:
             pip install --upgrade -e /usr/src/artman/
       - run:
           name: Build Pub/Sub (Java)
-          command: artman --local --config=/googleapis/google/pubsub/artman_pubsub.yaml publish --dry-run --github-username foo --github-token bar --target staging java_gapic
+          command: |
+            git clone https://github.com/googleapis/api-client-staging /tmp/api-client-staging
+            artman --local --config=/googleapis/google/pubsub/artman_pubsub.yaml publish --dry-run --local-repo-dir=/tmp/api-client-staging --github-username foo --github-token bar --target staging java_gapic
       - run:
           name: Build Speech (Java)
           command: artman --local --config=/googleapis/google/cloud/speech/artman_speech_v1.yaml publish --dry-run --github-username foo --github-token bar --target staging java_gapic
--- a/artman/cli/main.py
+++ b/artman/cli/main.py
@@ -234,7 +234,7 @@ def parse_args(*args):
         help='[Optional] When specified, artman will neither look up the '
         'publishing configuration in ~/.artman.config.yaml, nor clone a new '
         'github repo. Instead, it will use the specified directory to stage '
-        'generated result. This only work under dry run mode.', )
+        'generated result. This only works under --dry-run mode.', )
     parser_publish.set_defaults(dry_run=False)
 
     return parser.parse_args(args=args)
--- a/artman/cli/main.py
+++ b/artman/cli/main.py
@@ -231,10 +231,10 @@ def parse_args(*args):
         '--local-repo-dir',
         type=str,
         default=None,
-        help='[Optional] When specified, artman will skip the lookup local '
+        help='[Optional] When specified, artman will neither look up the '
         'publishing configuration in ~/.artman.config.yaml, nor clone a new '
         'github repo. Instead, it will use the specified directory to stage '
-        'generated result. This is only working when under dry run mode.', )
+        'generated result. This only work under dry run mode.', )
     parser_publish.set_defaults(dry_run=False)
 
     return parser.parse_args(args=args)
@@ -510,7 +510,7 @@ def _run_artman_in_docker(flags):
 
 
 def _change_owner(flags, pipeline_name, pipeline_kwargs):
-    """Change file/folder ownership if necessary."""
+    """Change file/directory ownership if necessary."""
     user_host_id = int(os.getenv('HOST_USER_ID', 0))
     group_host_id = int(os.getenv('HOST_GROUP_ID', 0))
     # When artman runs in Docker instance, all output files are by default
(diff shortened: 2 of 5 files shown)


artman/cli/main.py
```diff
@@ -227,6 +227,14 @@ def parse_args(*args):
         action='store_true',
         help='[Optional] When specified, artman will skip the remote '
         'publishing step.', )
+    parser_publish.add_argument(
+        '--local-repo-dir',
+        type=str,
+        default=None,
+        help='[Optional] When specified, artman will skip the lookup local '
```
vchudnov-g: s/skip the lookup local publishing/neither look up the publishing/
s/, nor/nor/

artman/cli/main.py
```diff
@@ -227,6 +227,14 @@ def parse_args(*args):
...
+        'publishing configuration in ~/.artman.config.yaml, nor clone a new '
+        'github repo. Instead, it will use the specified directory to stage '
+        'generated result. This is only working when under dry run mode.', )
```
vchudnov-g: s/This is only working when under /This only works under/

artman/cli/main.py
```diff
@@ -291,6 +299,14 @@ def normalize_flags(flags, user_config):
     pipeline_args['local_paths'] = support.parse_local_paths(
         user_config, flags.root_dir)
 
+    if flags.subcommand == 'publish' and flags.local_repo_dir:
+        if not flags.dry_run:
+            logger.error('`--dry-run` flag must be passed when '
+                         '`--local-repo-dir` is specified')
+            sys.exit(96)
```
vchudnov-g: why 96? :)

artman/cli/main.py
```diff
@@ -525,5 +542,16 @@ def _change_owner(flags, pipeline_name, pipeline_kwargs):
             os.chown(gapic_config_path, user_host_id, group_host_id)
 
 
+def _change_folder_owner(folder, user_host_id, group_host_id):
```
vchudnov-g: Just curious about your choice to use "folder" rather than "directory" in a bunch of places. Is this a convention? Otherwise I would stick to just one term, "directory"

.circleci/config.yml
```diff
@@ -189,7 +189,9 @@ jobs:
             pip install --upgrade -e /usr/src/artman/
       - run:
           name: Build Pub/Sub (Java)
-          command: artman --local --config=/googleapis/google/pubsub/artman_pubsub.yaml publish --dry-run --github-username foo --github-token bar --target staging java_gapic
+          command: |
+            git clone https://github.com/googleapis/api-client-staging /tmp/api-client-staging
+            artman --local --config=/googleapis/google/pubsub/artman_pubsub.yaml publish --dry-run --local-repo-dir=/tmp/api-client-staging --github-username foo --github-token bar --target staging java_gapic
```
vchudnov-g: I'm wondering whether it would be more informative to call the current semantics of `--dry_-run` something like `--no_push` instead, meaning we're not pushing any generated code to any place outside the local machine (regardless of what the publishing type is, GITHUB or something else in the future).

Then we could use `--dry-run` in its more conventional meaning: echoing the commands/processes/tasks artman would run, but without actually running them.

WDYT? Obviously it would be simpler to do this in two follow-up CLs, one for the flag rename and one for the new dry run functionality.

artman/cli/main.py
ethanbao: done

artman/cli/main.py
ethanbao: Done.

artman/cli/main.py
ethanbao: Artman convention :)

artman/cli/main.py
ethanbao: Done.

.circleci/config.yml
ethanbao: I saw your point. Currently `--dry-run` is a subflag of `publish` command. It means do everything except publish (or push). So it is equivalent to `--no_push`. I got the name from some internal usage, where a tool prepares the config (including merging and normalization) without real pushing. I think we won't have `--no-push` and `--dry-run` for two different behaviors. Between those two, I personally prefer the latter.

vchudnov-g: I guess what I'm trying to say is that I've seen `--dry-run` conventionally refer to printing the commands and configs that would be executed, so in this case it would print `git push ....` and implies a `--no_push`. So maybe you're right , we could do without `--no_push` if you make `--dry-run` specify the commands it's not running.

artman/cli/main.py
vchudnov-g: For all error messages? It would be nice to have different codes for different errors to make scripting easier.

artman/cli/main.py
```diff
@@ -227,6 +227,14 @@ def parse_args(*args):
...
         help='[Optional] When specified, artman will skip the remote '
         'publishing step.', )
+    parser_publish.add_argument(
+        '--local-repo-dir',
+        type=str,
+        default=None,
+        help='[Optional] When specified, artman will neither look up the '
+        'publishing configuration in ~/.artman.config.yaml, nor clone a new '
+        'github repo. Instead, it will use the specified directory to stage '
+        'generated result. This only work under dry run mode.', )
```
vchudnov-g: s/works/works/

vchudnov-g: s/dry run/--dry-run/

artman/tasks/publish/local.py
```diff
@@ -51,12 +52,16 @@ def execute(self, git_repo, local_paths, output_dir,
         if repo_name.endswith('.git'):
             repo_name = repo_name[:-4]
 
-        # Where is the target git repo located?
-        # Start by checking for an explicit path in `local_paths`, and then
-        # if none is found, clone the repo to output_dir.
+        # Artman will find the local repo dir in the following order:
```
vchudnov-g: "... dir via the following steps:"

artman/tasks/publish/local.py
```diff
@@ -51,12 +52,16 @@ def execute(self, git_repo, local_paths, output_dir,
...
+        # 1. Check whether an explicit `--local_repo_dir` flag is passed.
+        # 2. Check whether a matched local repo dir is specified in user config.
+        # 3. Clones the repo to output_dir, and use the cloned repo dir.
```
vchudnov-g: s/Clones/Otherwise, clone/

artman/cli/main.py
ethanbao: The current convention is a bad practice. I will follow up with a PR to categorize those errors and return different error code.

.circleci/config.yml
ethanbao: Sounds good. I am open to flag name change.

artman/cli/main.py
ethanbao: Done

ethanbao: Done

artman/tasks/publish/local.py
ethanbao: Done.

PR #366, opened 2018-01-08
Exit artman with non-zero code  when execution failed.

ethanbao:


artman/cli/main.py
```diff
@@ -69,7 +69,8 @@ def main(*args):
                 pipeline.flow, engine='serial', store=pipeline.kwargs)
             engine.run()
         except:
-            logger.fatal(traceback.format_exc())
+            logger.error(traceback.format_exc())
+            sys.exit(128)
```
geigerj: Error exit codes are [recommended](https://docs.python.org/2/library/sys.html#sys.exit) to be in the 1-127 range.

ethanbao: Good catch. Fixed.

PR #367, opened 2018-01-09
 Introduce --toolkit-dir to override toolkit directory during coding

vchudnov-g:
This allows running the artman command while incorporating changes to a local clone of the googleapis/toolkit repository.

### DIFF
--- a/artman/cli/main.py
+++ b/artman/cli/main.py
@@ -264,7 +264,8 @@ def normalize_flags(flags, user_config):
     """Combine the argparse flags and user configuration together.
 
     Args:
-        flags (argparse.Namespace): The flags parsed from sys.argv
+        flags (argparse.Namespace): The flags parsed from sys.argv, which may
+                                    be mutated in this function.
         user_config (dict): The user configuration taken from
                             ~/.artman/config.yaml.
 
--- a/artman/cli/main.py
+++ b/artman/cli/main.py
@@ -153,6 +153,15 @@ def parse_args(*args):
         'will be raised with instructions.',
     )
     parser.add_argument(
+        '--toolkit-dir',
+        type=str,
+        default='',
+        help='[Optional] Local directory containing the local clone of the '
+        'github/toolkit repo to use, instead of the one specified in the '
+        'configuration files. This flag is useful for testing development '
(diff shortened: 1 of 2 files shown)


artman/cli/main.py
```diff
@@ -152,6 +152,15 @@ def parse_args(*args):
...
     )
+    parser.add_argument(
+        '--toolkit-dir',
+        type=str,
+        default='',
+        help='[Optional] Local directory containing the local clone of the '
+        'github/toolkit repo to use, instead of the one specified in the '
+        'configuration files. This flag is useful for testing development '
+        'changes to toolkit via an artman invocation.',
+    )
```
ethanbao: Were you suggesting to mount the local toolkit to the docker container and use that inside the docker container?

vchudnov-g: Yes, exactly. I can mount via "-v .../toolkit/:/tmp/toolkit" to Docker and then run artman inside it with "--toolkit-dir /tmp/toolkit"

vchudnov-g:
BTW, @ethanbao , the CI error message `ImportError: No module named 'msgpack'` appears to be unrelated to my changes. Any ideas what could be causing it?

ethanbao:
msgpack error should have been fixed by https://github.com/googleapis/artman/commit/d637f72d1b928b803bb652978364544b510a04c0#diff-b4ef698db8ca845e5845c4618278f29a

Fetch the latest repo and that error should be gone.

artman/cli/main.py
ethanbao: In ~/.artman/config.yaml, there is local_paths.toolkit which will point artman to local toolkit dir. But that only works for local mode. If you are okay with mount via -v, you can simply update the ~/.artman/config.yaml inside the docker image without introducing this new flag.

Unlike --local-repo-dir that was recently added, I feel this --toolkit-dir will be even less used by normal users.

vchudnov-g: Ah, OK. I couldn't this to work by changing the toolkit path in my non-Docker .artman.config.yaml. But changing it inside the Docker container works.
(We may want to think about propagating that information in automatically.)
I agree, in that case this PR is not necessary. I'll close it.

PR #369, opened 2018-01-09
bump googleapis version

neozwu:


neozwu:
@ethanbao Done. PTAL.

PR #370, opened 2018-01-09
fix golden test

neozwu:


ethanbao:
The golden-test will fail. Ignore the failure and merge the PR.

PR #371, opened 2018-01-09
Validate artman yaml format before the execution

ethanbao:


artman/cli/main.py
```diff
@@ -46,6 +48,13 @@
 ARTMAN_DOCKER_IMAGE = 'googleapis/artman:%s' % VERSION
 RUNNING_IN_ARTMAN_DOCKER_TOKEN = 'RUNNING_IN_ARTMAN_DOCKER'
 
+# Error messages
+CONFIG_NOT_FOUND_ERROR_MESSAGE_FORMAT = (
+    'Artman YAML cannot be found at `%s`. Please check the file location.')
+INVALID_CONFIG_ERROR_MESSAGE_FORMAT = (
+    'Artman YAML %s is invalid. Please pass a valid Artman YAML (e.g. '
```
geigerj: I would rather print the cause of the parsing error (line number etc.), assuming the yaml library can provide this. I'd prefer not to link to the master branch of a GitHub repository in the error message, which is brittle if we refactor googleapis in the future, etc.

artman/cli/main.py
```diff
@@ -286,6 +295,12 @@ def normalize_flags(flags, user_config):
         flags.config = os.path.join(flags.root_dir, flags.config)
     else:
         flags.config = os.path.abspath(flags.config)
+    try:
+        _validate_config_flag(flags.config)
```
geigerj: It looks like this parses the config, and then throws it away. Why are we parsing it twice?

geigerj: It looks like this parses the config, and then throws it away. Why are we parsing it twice?

artman/cli/main.py
ethanbao: Removed. Added a TODO to point to a formal artman config page link once we have one.

artman/cli/main.py
ethanbao: Good point. Moved the logic to loader.py which is a better place to do the validation under current implementation.

artman/config/loader.py
```diff
@@ -24,6 +24,14 @@
...
 
 from artman.config.proto.config_pb2 import Artifact, Config
+from artman.utils.logger import logger
+
+# Error messages
+CONFIG_NOT_FOUND_ERROR_MESSAGE_FORMAT = (
+    'Artman YAML cannot be found at `%s`. Please check the file location.')
+# TODO(ethanbao): Add a reference link once Artman schema spec is publicly
+# available.
+INVALID_CONFIG_ERROR_MESSAGE_FORMAT = 'Artman YAML %s is invalid.'
```
geigerj: Is there no way to get an error message or a line number out of the parser?

ethanbao: We lose the track in the mid. Because protobuf python library only support validation of a json string not yaml string, I ended up converting yaml string to json string first, the line number mapping is not retained.

If the yaml file itself is in a invalid format (without checking it against a protobuf message type schema), the line number is returned as part of yaml.parser.ParserError,

artman/config/loader.py
```diff
@@ -56,12 +64,19 @@ def _read_artman_config(artman_yaml_path):
...
+    try:
+        with open(artman_yaml_path, 'r') as f:
+            # Convert yaml into json file as protobuf python load support paring of
+            # protobuf in json or text format, not yaml.
+            artman_config_json_string = json.dumps(yaml.load(f))
+        config_pb = Config()
+        json_format.Parse(artman_config_json_string, config_pb)
+    except (json_format.ParseError, yaml.parser.ParserError) as e:
+        logger.error(INVALID_CONFIG_ERROR_MESSAGE_FORMAT % artman_yaml_path)
+        raise ValueError(str(e))
```
geigerj: Why not just `raise`?

artman/config/loader.py
geigerj: I see. I hope the `json_format.ParseError` gives enough information to the user to debug if the yaml is valid but the instance of the proto is not.

artman/config/loader.py
ethanbao: If the artman yaml input goes wrong, it is most likely cased by json_format.ParseError. Then it could cause confusing to raise a json format error to the caller while the input is a yaml file.

OTOH, I would agree that ValueError is not good either. If artman is going to be used as a library not a CLI, I would consider create some custom error type.

geigerj: I am more concerned that you are discarding both the stack trace and any additional attributes on the original exception when you create a new error and stringify the old error. That can potentially make debugging much more difficult.

ethanbao: I would say the stack trace in this particular case is not quite helpful to the artman users. Below is an example error output under current implementation.

$artman --config google/cloud/speech/v1/cloud_speech_gapic.yaml generate java_gapic
artman> Artman YAML google/cloud/speech/v1/cloud_speech_gapic.yaml is invalid.
artman> Artifact config loading failed with `Message type "googleapis.artman.Config" has no field named "interfaces".
 Available Fields(except extensions): <MessageFields sequence>`

I agree that the stack trace would be useful for some cases. I logged the trace information at debug logging level. When one pass --verbose, they will see the full stacktrace.

ethanbao: PTAL, @geigerj

geigerj: I still don't understand the advantage of throwing away the original error here. If anything else goes wrong, it's useful to have the stacktrace, and it's still seems to be me this is throwing away a potentially rich exception object and replace it with its string representation for no real benefit.

ethanbao: What is the value of printing the stacktrace to sys.err if I am a just regular artman user? Doesn't `Message type "googleapis.artman.Config" has no field named "interfaces"` already tell everything artman could return? Or I might misread your comment. Do you want such exceptions be captured and handled by artman, you would suggest just raise it so that artman execution will exit with a full stacktrace?

geigerj: > What is the value of printing the stacktrace to sys.err if I am a just regular artman user? 

It lets you set breakpoints if you're using a debugger, for one. These exceptions are also objects, so you can catch and inspect them normally, but once you throw away the original object, there's no way to retrieve any attributes that might not be captured in its string representation.

FWIW, this is solved in Python 3 with the `raise ... from ...` syntax.

PR #377, opened 2018-01-11
Run smoketest in verbose mode and upload log to circleci artifacts when it fails

ethanbao:
Note while docker-based circleci does support running docker in docker, the folder mounting works differently. As a result, the log doesn't upload to artifacts. To compensate that, we now print the full log inside the docker when the test fails.


ethanbao:
ping @geigerj @lukesneeringer

geigerj:
@ethanbao For context, what's currently preventing `/tmp/reports` from being uploaded?

ethanbao:
The artifact can be uploaded. The problem is that the log (file change) in the container is not passed to the host. This seems to be a very specific CircleCI behavior as my local running using the same command could get the log data. Because CircleCI runs test in docker, making my current test config effectively docker in docker. I guess that is why the mounting works differently.

geigerj:
I'm still a little confused by this.

The logger logs to `log_file` while generating the GAPICs. Then, on failure, it reads the logs that it just wrote, and re-logs them?

ethanbao:
When the execution failed, the test (which run inside the container) reads the log file inside the container it print it to the console (which is available in the CI report). I have to do that because the host (which runs `docker` command) cannot see the log file which is written inside the container, and therefore nothing will be uploaded as artifact (note: there is no problem when running the smoketest on a regular dev machine). I bet this is because the folder mounting doesn't really work when running docker inside docker.

test/smoketest_artman.py
```diff
@@ -80,6 +80,8 @@ def run_smoke_test(root_dir, log):
     for msg in failure:
         logger.error(msg)
     if failure:
+        with open(log_file, 'r') as logs:
+            print(logs.read())
```
geigerj: Suggest a comment explaining why we need a print statement, since I don't think it's immediately obvious.

test/smoketest_artman.py
```diff
@@ -137,4 +131,16 @@ def _setup_logger(log_file):
 if __name__ == '__main__':
     flags = parse_args(*sys.argv[1:])
 
-    run_smoke_test(os.path.abspath(flags.root_dir), os.path.abspath(flags.log))
+    try:
```
geigerj: This can be simplified:

```
try:
  run_smoke_test(...)
except:
  sys.exit(...)
finally:
  with open(...) as logs:
    print(logs.read())

```

Or:

```
def run_smoke_test(...):
  ...
  if failure:
    logger.error('Smoke test failed')

...

if __name__ == '__main__':
  try:
    run_smoke_test(...)
  finally:
    with open(...) as logs:
      print(logs.read())
```

ethanbao: The smoketest would need to exit with non-zero status code when it fails.

geigerj: For nonzero exit status:

```
def run_smoke_test(...):
  ...
  assert(not failure)

...

if __name__ == '__main__':
  try:
    run_smoke_test(...)
  finally:
    with open(...) as logs:
      print(logs.read())
```

test/smoketest_artman.py
```diff
@@ -15,7 +15,6 @@
 # limitations under the License.
 
 """Artman smoke tests.
```
geigerj: revert this file?

geigerj:
Please update PR description/commit title before squashing.

PR #380, opened 2018-01-18
Whitelist several artifacts that fails the smoketest.

ethanbao:
Note: these are existing failure that was just recently uncovered because a bug was fixed that will let toolkit return non-zero status code when generation fails.


ethanbao:
I realized that some errors was caused by the fact that a new gapic config gets generated and replaces the one in github before library generation. This introduces both false positive and false negative. Skip the gapic config generation in the smoketest. If you think it is valuable to include gapic config generation. I will add a separate test for that.

geigerj:
@ethanbao I think a separate test will be useful. I am surprised GAPIC config generation ever fails, since there's no additional user configuration needed for it beyond the IDL.

ethanbao:
It failed for bigtable for some unknown reason. It works for other APIs but it will replace the existing gapic yaml with newly-generated one. And there is no guarantee that the auto-generated one will always work. More importantly, it hides problem, making smoketest less useful.

geigerj:
Understood; for the GAPIC library generation smoke test, we should use the checked-in configs, not fresh ones.

ethanbao:
I made a change to move the gapic_config generation (if any) after the library generation.

test/smoketest_artman.py
```diff
@@ -51,13 +52,27 @@ def run_smoke_test(root_dir, log):
...
                                       recursive=True):
         artman_config = _parse(artman_yaml_path)
+        # Reorder the artifacts to move the gapic_config generation to the end.
+        # Otherwise, the following GAPIC library generation is going to use the
+        # newly-generated GAPIC config instead of the one in github, leading to
+        # false negative and false positive.
+        artifacts = []
+        gapic_config_artifact = None
         for artifact in artman_config.artifacts:
+            if config_pb2.Artifact.GAPIC_CONFIG == artifact.type:
```
geigerj: nit: in case of multiple `GAPIC_CONFIG` artifacts:

`artifacts = sorted(artman_config.artifacts, key=lambda a: config_pb2.Artifact.GAPIC_CONFIG == a.type))`

?

ethanbao: Super nice

PR #385, opened 2018-02-01
Java: update generated 'gapic_code_dir' value to include api_version.

vam-google:
Also update prefix from from 'google-cloud-' to 'gapic-google-cloud-' to make it consistent with directory names in api-client-staging.

Without this, in case of subsequent local generation (calling 'artman generate') of two different versions for same client, one overrides another, which makes impossible to locally generate multiple versions of same clients (for example language-v1 and language-v1beta2).

Tested that local generation and publishing works fine after the change.


ethanbao:
Please fix the unit tests and golden test (https://github.com/googleapis/artman/blob/master/test/golden/library_example.golden).

vam-google:
@ethanbao, updated, PTAL

ethanbao:
LGTM

PR #387, opened 2018-02-07
Always run golden test during presubmit and run smoketest conditionally.

ethanbao:
The golden test (including the new docker image building using cache) took about 9 mins (e.g. https://circleci.com/gh/ethanbao/artman/2960)

Running golden test in this way will help capture artman output change during the PR presubmit.


.circleci/config.yml
```diff
@@ -45,10 +45,15 @@ jobs:
       - run:
           name: Run smoketests
           command: |
-            docker run -it -e RUNNING_IN_ARTMAN_DOCKER=True --name smoketest artman /bin/bash -c "git clone https://github.com/googleapis/googleapis /tmp/googleapis_smoketest; python3 /artman/test/smoketest_artman.py --root-dir=/tmp/googleapis_smoketest --log=/tmp/smoketest.log" || OUT=$?
-            docker cp smoketest:/tmp/smoketest.log /tmp/smoketest.log
-            if [ $OUT -ne 0 ];then
-              exit $OUT
+            if [ "${RUN_SMOKETEST}"]; then
```
geigerj: nit: space before `]`

.circleci/config.yml
```diff
@@ -316,6 +321,8 @@ workflows:
       - smoke-php
       - smoke-python
       - smoke-ruby
+      # TODO(ethanbao): Separate the test and docker release into two jobs.
+      - build_and_push_docker_image
```
geigerj: Why is this being done twice? Also, without a branch filter like below, does this mean that docker images might get pushed from arbitrary branches?

ethanbao: Good catch. The docker push should only happen for master branch.

ethanbao:
For googleapis/artman CircleCI project, I have set this `RUN_SMOKETEST` env variable. For the forked repos, developers can opt to set the env variable as well, but that will slow down the tests.

The purpose of this change is to add golden test back (under `build_and_push_docker` name though). The way we used to smoketest artman is problematic because it doesn't rebuild the Docker (therefore won't pick up the latest shared config change, which will affect the artman output). What happened is the PR gets merged with all tests passing, while one future PR will receive a confusing golden test failure report which is due to an earlier PR.

geigerj:
> For the forked repos, developers can opt to set the env variable as well, but that will slow down the tests.

I wouldn't worry about this. I doubt most forks care about testing or accept PRs.

.circleci/config.yml
```diff
@@ -45,11 +45,55 @@ jobs:
       - run:
           name: Run smoketests
           command: |
-            docker run -it -e RUNNING_IN_ARTMAN_DOCKER=True --name smoketest artman /bin/bash -c "git clone https://github.com/googleapis/googleapis /tmp/googleapis_smoketest; python3 /artman/test/smoketest_artman.py --root-dir=/tmp/googleapis_smoketest --log=/tmp/smoketest.log" || OUT=$?
-            docker cp smoketest:/tmp/smoketest.log /tmp/smoketest.log
-            if [ $OUT -ne 0 ];then
-              exit $OUT
+            if [ "${RUN_SMOKETEST}" ]; then
```
geigerj: I still don't know why this needs to be flag-guarded, but won't block the review over it.

.circleci/config.yml
```diff
@@ -45,11 +45,55 @@ jobs:
...
+
+  build_and_release_docker_image:
+    working_directory: /usr/src/artman
+    docker:
+      - image: docker:17.05.0-ce-git
+    steps:
+      - checkout
+      - setup_remote_docker
+      - run:
+          name: Fingers crossed!
```
geigerj: Why do we have these `echo` commands?

.circleci/config.yml
```diff
@@ -45,11 +45,55 @@ jobs:
...
+          command: |
+            echo "Fingers crossed!"
+      - restore_cache:
```
geigerj: Is there a way to factor out this set-up? If unknown, a TODO?

.circleci/config.yml
ethanbao: The whole smoketest took ~1 hour, I don't want to have it running during each PR review, but want to have it running automatically when the PR gets merged. I didn't figure out other way except using the circleci project env variable.

.circleci/config.yml
ethanbao: For fun

.circleci/config.yml
geigerj: >  I don't want to have it running during each PR review, but want to have it running automatically when the PR gets merged.

Doesn't a branch filter on `master` work for this?

.circleci/config.yml
ethanbao: `restore_cache` is a circleci built-in task. What do you mean by factor out?

.circleci/config.yml
geigerj: FWIW, the way acceptance tests are set up in Ruby is that they run only if `CIRCLE_BRANCH` == `master` and `CIRCLE_PR_NUMBER` is empty.

.circleci/config.yml
geigerj: I mean that lines 74-96 duplicate lines 15-37.

.circleci/config.yml
ethanbao: Probably will work in practice, as the PR is most likely from a non-master forked branch. But PR like this (https://github.com/googleapis/artman/pull/386) will trigger the full smoketest.

ethanbao:
PTAL @geigerj

PR #388, opened 2018-02-07
Introduce `skip_proto_gen` and `skip_grpc_gen` flags in artman yaml

neozwu:
Introduce `skip_proto_gen` and `skip_grpc_gen` flags to toggle proto and grpc code gen tasks in GapicClientPipeline (currently, for some languages, running gapic task will always run proto and grpc tasks as well). These flags will be useful for java longrunning, where only gapic but not proto/grpc artifact is needed.


neozwu:
close this PR since the function is now already supported.

PR #389, opened 2018-02-10
Refactor artman user config

ethanbao:
Improve the artman usability by a) removing artman user config fields that are no longer necessary, b) proto-fy artman user config and c) making artman user config (~/.artman/config.yaml) optional.

**Context**:
Artman user config was originally introduced in artman PR#159. The PR largely improved the usability of artman at that time, and many fields were named to be consistent with artman CLI by that time (e.g. reporoot). Another refactoring (PR#254) was made towards artman config (not user config) and artman CLI in Aug 2017, which made some existing artman user config fields no longer necessary.

Below are the proposed changes:
* Changes to artman user config schema:

Current | Proposed | Description
-- | -- | --
`local_paths.reporoot` | Remove | This is a very legacy field. When it is specified, it tells artman the default value for local_paths.reporoot.[googleapis\|toolkit\|artman] user config fields, which then can be overridden by the next three fields.
`local_paths.reporoot.googleapis` | Remove, leaving `--root-dir` flag the only way to specify googleapis dir. | Specify the location of local `googleapis` dir.
`local_paths.reporoot.toolkit` | Keep but rename as `local.toolkit`. | Specify the location of local `toolkit` source code dir or binary jar. This is useful for local debugging.
`local_paths.reporoot.artman` | Remove | Specify the local artman source code. The idiomatic way to test local artman source code is to create a python virtualenv and install the artman from local source code.
`github.username` | Keep, which can be overridden by `--github_username` flag. | Specify the github username which is currently needed by `artman publish`
`github.token` | Keep, which can be overridden by `--github_token` flag. | Specify the github token which is currently needed by `artman publish`
`publish` | Removed, which is now replaced by combination of generate/publish actions with and without `--dry-run` flag. | Specify the default publishing policy.

* With the proposed change, the artman user config becomes optional. While artman user config definitely has its value, it could improve the getting-started experience if it becomes optional (when the physical ~/.artman/config.yaml doesn’t exist).

* When `--repo-root` flag is not specified, the current working directory is treated as the repo root for input files (e.g. protos, yamls).

A new protobuf file will be created for artman user config schema.


artman/cli/configure.py
```diff
@@ -46,107 +51,116 @@ def configure(log_level=logging.INFO):
...
-    reporoot = answer['reporoot']
-    for dep in ('api-client-staging', 'googleapis', 'toolkit'):
+    answer = LocalConfig()
+
+    # Ask the user for a local toolkit location.
+    while not answer.toolkit:
         location = six.moves.input(
-            'Path for {0} (default: {1}/{0}): '.format(dep, reporoot)
-        ).rstrip('/').strip()
+            'Where is your local toolkit? Please provide an absolute path: ')
```
geigerj: A new user might not know what "your local toolkit" refers to. Suggest "... your local toolkit repository? (If you do not have a toolkit repository, clone https://github.com/googleapis/toolkit/. Please provide an absolute path: "

artman/cli/configure.py
```diff
@@ -46,107 +51,116 @@ def configure(log_level=logging.INFO):
...
-    while not answer.get('token'):
-        answer['token'] = getpass.getpass('GitHub token (input is hidden): ')
+    while not answer.username:
+        answer.username = six.moves.input('GitHub username: ')
+    while not answer.token:
+        answer.token = getpass.getpass('GitHub token (input is hidden): ')
+
     return answer
+
+def _convert_json(d):
```
geigerj: Why do we need to covert proto names to snake_case?

artman/cli/configure.py
```diff
@@ -46,107 +51,116 @@ def configure(log_level=logging.INFO):
...
+        node_key = dumper.represent_data(item_key)
+        node_value = dumper.represent_data(item_value)
+
+        value.append((node_key, node_value))
+
+    return yaml.nodes.MappingNode(u'tag:yaml.org,2002:map', value)
+
+
+def _order_dict(od):
+    # The whole key order is flattened which is okay for artman user config
```
geigerj: > The whole key order is flattened

What does this mean?

artman/cli/support.py
```diff
@@ -62,54 +62,17 @@ def parse_github_credentials(config, argv_flags):
...
-        'username': argv_flags.github_username or config['username'],
-        'token': argv_flags.github_token or config['token'],
+        'username': argv_flags.github_username or github_config.username,
+        'token': argv_flags.github_token or github_config.token,
     }
 
 
-def parse_local_paths(user_config, root_dir):
-    """Parse all relevant local flags, given appropriate user config and user.
+def resolve(name, flags, default=None):
```
geigerj: ISTM this method is no longer necessary and can be replaced with `flags.vars().get(name, default)`

artman/config/loader.py
```diff
@@ -32,6 +34,7 @@
 # TODO(ethanbao): Add a reference link once Artman schema spec is publicly
 # available.
 INVALID_CONFIG_ERROR_MESSAGE_FORMAT = 'Artman YAML %s is invalid.'
+INVALID_USER_CONFIG_ERROR_MESSAGE_FORMAT = 'Artman YAML %s is invalid.'
```
geigerj: Is this repeated error message intentional? We should have different names for the Artman config for an API and the Artman user config.

artman/cli/configure.py
ethanbao: Accepted the suggestion.

artman/cli/configure.py
ethanbao: Good catch. Not necessary.

artman/cli/configure.py
ethanbao: Say an ideal artman user config is in the following field order

"""
field1:
  field1_a:
  field1_b:
  field_c:

field2:
  field2_a:
  field2_b:
  field_c:
"""

The order is defined as "field1, field1_a, field1_b, field2, field2_a, field2_b, field_c" in a flattend manner (vs nested). Note: there is the limitation to its expression as when `field_c` (field name) is defined inside multiple parent fields, it might not end up working. But this won't be a problem in practice.

geigerj: If you intend users to define the data in a flattened manner, why not define the proto in a flattened manner?

ethanbao: While artman only has 3 fields, it is reasonable to have everything flattened without impacting understand-ability. For the long run (if any), it will be more flexible using a structured configuration. That is my personal preference though.

geigerj: Either way is fine, but I do not understand why there would be a discrepancy between the proto and the YAML.

ethanbao: I did mean to make these fields consistent between proto definition and yaml output. WIthout this ordering implementation, after artman user run `configure-artman` it will end up with the following:

"""
github:
  token: token
  username: test
local:
  toolkit: /toolkit
"""

I would prefer `local` be in front of `github` though it is okay to change. It would be awkward to have `token` in front of `username`.

geigerj: Sorry, I don't follow. What does ordering have to do with flattened vs. nested?

ethanbao: It is confusing to call it flattening. Updated the comment.

geigerj: Thanks, I think I understand now: what is flattened is the ordering relation you've defined, not the data structure.

artman/config/proto/user_config.proto
```diff
@@ -0,0 +1,56 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package googleapis.artman;
+
+// Artman user config is used to store user settings. All configuration are
```
geigerj: "Artman user config" -> "The Artman user config"
"All configuration are" -> "All configuration is"

artman/config/proto/user_config.proto
```diff
@@ -0,0 +1,56 @@
...
+// optional unless being explicitly marked required. Having some configuration
```
geigerj: "unless being" -> "unless it is"
"Having some configuration will simplify" -> "Configuration can simplify"

artman/config/proto/user_config.proto
```diff
@@ -0,0 +1,56 @@
...
+// will simplify the CLI invocation (e.g. github config), or change the
+// execution behavior (e.g. local.toolkit).
```
geigerj: > change the execution behavior (e.g. local.toolkit)

What does this mean? if `local.toolkit` is specified, it doesn't cause artman always to run in local mode, right?

artman/config/proto/user_config.proto
ethanbao: Updated

artman/config/proto/user_config.proto
ethanbao: Updated

artman/config/proto/user_config.proto
ethanbao: Added clarification.

artman/config/proto/user_config.proto
```diff
@@ -16,10 +16,12 @@ syntax = "proto3";
...
 package googleapis.artman;
 
-// Artman user config is used to store user settings. All configuration are
-// optional unless being explicitly marked required. Having some configuration
-// will simplify the CLI invocation (e.g. github config), or change the
-// execution behavior (e.g. local.toolkit).
+// The Artman user config is used to store user settings. All configuration is
+// optional unless it is explicitly marked as required. Configuration can
+// simplify the CLI invocation (e.g. github config), or change the behavior.
+// For example, when `local.toolkit` is configured to a local toolkit path, it
```
geigerj: suggest: "For example, if `local.toolkit`is specified, artman will run toolkit from a local source repository when executed in local mode."

artman/config/proto/user_config.proto
```diff
@@ -0,0 +1,62 @@
...
+  // used. This is useful for toolkit debugging. This configuration currently
+  // only work when running in artman local mode.
+  //
+  // Note: Absolute path should be used.
+  string toolkit = 1;
+}
+
+message GitHubConfig {
+  // The GitHub username, which is used to submit Gitcommit and create GitHub
+  // PR. Artman user can pass a different value when invoking `artman` CLI via
```
geigerj: nit: "Artman user" -> "The user"

artman/config/proto/user_config.proto
```diff
@@ -0,0 +1,62 @@
...
+  // `--github-username` flag.
+  string username = 1;
+
+  // The GitHub personal access token that matches the specified GitHub
+  // username. Artman user can pass a different value when invoking `artman` CLI
```
geigerj: nit: "Artman user" -> "The user"

PR #390, opened 2018-02-21
Temporarily whitelist php_gapic generation of videointelligence_v1

ethanbao:
Temporarily whitelist php_gapic generation of videointelligence_v1 in smoketest.


test/smoketest_artman.py
```diff
@@ -42,6 +42,8 @@
     'gapic_config@google/bigtable/admin/artman_bigtableadmin.yaml',
     'java_gapic@google/cloud/dialogflow/artman_dialogflow_v2beta1_java.yaml',
```
garrettjonesgoogle: Wait, why is the java one whitelisted?

ethanbao:
Removed that artifact. It was listed because it didn't work earlier.

garrettjonesgoogle:
The Java LGTM

PR #399, opened 2018-03-01
Use java resource names plugin v 0.0.7

andreamlin:
New protoc java resource names plugin at https://pypi.org/project/protoc-java-resource-names-plugin/


andreamlin:
Yep, api-client-staging is updated (https://github.com/googleapis/api-client-staging/pull/603).

@ethanbao  PTAL

PR #401, opened 2018-03-19
Update gRPC version.

lukesneeringer:
This allows `option php_namespace` to be recognized.


michaelbausor:
Do we know why this has turned up as a problem now, rather than much earlier? There are many APIs in googleapis that have the php_namespace option, and have done for several months.

lukesneeringer:
> Do we know why this has turned up as a problem now, rather than much earlier? There are many APIs in googleapis that have the php_namespace option, and have done for several months.

My guess is that we never tried to build any of them in Python or Ruby. :-)

michaelbausor:
> My guess is that we never tried to build any of them in Python or Ruby. :-)

That could be true... but Spanner, Pubsub, Logging, Firestore, Monitoring, Error reporting, Video Intelligence and Trace have all had this option for 4 months. There are also the new APIs like DataTransfer, which have had it almost since the proto was open sourced.

lukesneeringer:
It is possible, although I would _expect_ this to affect artman.
Let me double-check this fix. :-)

michaelbausor:
I agree with Luke's assessment, that I would expect this to affect artman as well, although I am not positive.

I would also note that artman uses protoc/protobuf in some funny ways. For example, the function here gets the protobuf path using a gradle task in toolkit - I can see there being a strange interaction between having different versions of protoc CLI and toolkit gradle protobuf library.
https://github.com/googleapis/artman/blob/master/artman/utils/protoc_utils.py#L341

lukesneeringer:
> I agree with Luke's assessment, that I would expect this to affect artman as well, although I am not positive.

We just did testing on our side, and with the PR it works. I think this should belong in one of the great mysteries of the universe (at least for now).

> I would also note that artman uses protoc/protobuf in some funny ways. For example, the function here gets the protobuf path using a gradle task in toolkit - I can see there being a strange interaction between having different versions of protoc CLI and toolkit gradle protobuf library.

It also uses grpcio-tools and grpc-tools in Python and Ruby, respectively, which I am fairly certain bundle _their own_ protobuf versions.

PR #409, opened 2018-04-02
Enable Discogapic gen

andreamlin:
Enable generation of client libraries and config files from Discovery docs.

To generate the Java GCE client, run
`artman --config google/compute/artman_compute.yaml --local generate java_discogapic`

To generate the default config compute_gapic.yaml, run
`artman --config google/compute/artman_compute.yaml --local generate discogapic_config`

^ The above commands assume https://github.com/googleapis/googleapis/pull/498 is patched in locally.


andreamlin:
PTAL

andreamlin:
There's a fair bit of ctrl+c, ctrl+v going on in this diff. Open to suggestions for parameterizing gapic/config gen functions on parameters of input sources.

garrettjonesgoogle:
@vam-google could you review this?

test/pipelines/test_batch_generation.py
```diff
@@ -107,8 +128,8 @@ def test_get_language_api_flows(self):
                 _, kw = make.call_args
                 print('kw:', kw)
                 apis.append(kw['api_name'])
-            assert make.call_count == 2
-            assert apis == ['logging', 'pubsub']
+            assert make.call_count == 3
+            assert apis == ['logging', 'pubsub', 'simplecompute']
```
lukesneeringer: Why is this called "_simple_ compute"?

artman/tasks/gapic_tasks.py
```diff
@@ -112,6 +138,35 @@ def validate(self):
...
+        gapic_api_yaml, discogapic_language_yaml, package_metadata_yaml,
+        gapic_code_dir, api_name, api_version, organization_name):
+        existing = glob.glob('%s/*' % gapic_code_dir)
+        if existing:
+            self.exec_command(['rm', '-r'] + existing)
+        gapic_yaml = gapic_api_yaml + discogapic_language_yaml
+        gapic_args = ['--gapic_yaml=' + os.path.abspath(yaml)
+                      for yaml in gapic_yaml]
+        args = [
+            # TODO(andrealin): Get right absolute path for discovery_doc
```
lukesneeringer: This todo might be done? (Not obvious to me.)

artman/tasks/gapic_tasks.py
```diff
@@ -51,6 +51,32 @@ def validate(self):
...
+        output_dir, api_name, api_version, organization_name):
+        api_full_name = task_utils.api_full_name(
+            api_name, api_version, organization_name)
+        config_gen_dir = os.path.join(
+            output_dir, api_full_name + '-config-gen')
+        self.exec_command(['mkdir', '-p', config_gen_dir])
+        config_gen_path = os.path.join(config_gen_dir,
+                                       api_full_name + '_gapic.yaml')
+        args = [
+            '--discovery_doc=' + os.path.abspath(discovery_doc),
```
lukesneeringer: You probably want an explicit `os.path.expanduser`, because `abspath` and `realpath` do not resolve `~` (sadly).

```python
os.path.abspath(os.path.expanduser(discovery_doc))
```

andreamlin: Done.

artman/tasks/gapic_tasks.py
andreamlin: I'm not sure why this doesn't work either :(
I just know that in googleapis/.../artman_compute.yaml, I have to set
`discovery_doc: google/compute/v1/compute.v1.json`
even though I would prefer, in line with convention, to set
`discovery_doc: v1/compute.v1.json`

test/pipelines/test_batch_generation.py
andreamlin: I had it at "simplecompute" because it wasn't the real, full compute API.
Changed it back to "compute" because these are just test resource files and I think other people understand it's not the real thing.

PR #411, opened 2018-04-11
Discogapic: fetch Discovery document from remote source

andreamlin:
Enable fetching the Discovery document from github or from a URL (in addition to having a locally sourced option).

Changes config.proto's `discovery_doc` from a string filepath into a message type that includes the filepath and a type (GITHUB, LOCAL, etc).


andreamlin:
@garrettjonesgoogle There's been some discussion between @vam-google and me about whether this is a useful PR. On one hand, it saves the user the trouble of manually checking out the latest discovery doc, but on the other hand, it does add some complexity from the new config and from possibly requiring a remote connection to github. Thoughts?

garrettjonesgoogle:
I think we should maintain parity between usage of googleapis/googleapis and discovery-artifact-manager. Checkouts from github might be nice, but I would prefer to do a design that considers what workflows we want to support. So, until we have that higher-level discussion, I might want to hold off on this feature.

andreamlin:
Ok, putting this feature on hold.

PR #412, opened 2018-04-13
Enable discogapic to be run from discovery-artifact-manager

andreamlin:


andreamlin:
PTAL

artman/pipelines/gapic_generation.py
```diff
@@ -267,6 +266,7 @@ def _get_gapic_codegen_tasks(self, language, **kwargs):
             list: A list of Task subclasses.
         """
         return [
+            tasks.io.MakeOutputDirectoryTask,
```
vam-google: How did it work before (there was no specific directory-maker before for gapic tasks but they still worked)? Also this task looks misplaced (should it be used in a similar way as `PrepareGoogleapisDirTask`?). Also it is named too specifically (it does only one operation and is named to reflect that, but could be something more generic like `PrepareOutputDirectoryTask` or something, which would allow modifying it without making its name wrong). In general if we can avoid having this task we should try not having it.

PR #413, opened 2018-04-20
Update toolkit and googleapis in artman, fix golden tests

vam-google:


neozwu:
@lukesneeringer PTAL. A customer is blocked by this.

PR #417, opened 2018-05-11
update java resource plugin and bump for release

pongad:


pongad:
Let's not merge this yet. We should also update toolkit.

jadekler:
> Let's not merge this yet. We should also update toolkit.

Any idea of an ETA?

pongad:
@jadekler I hope to merge this in order or days.

lukesneeringer:
Node seems to fail to build again. Looking into it. Argh.

lukesneeringer:
@andreamlin Good catch.

lukesneeringer:
This can be merged after the tests (eventually) pass. Then it is a PyPI release and Docker upload, in that order.

PR #418, opened 2018-05-11
Stripping down package metadata file

garrettjonesgoogle:
This corresponds to https://github.com/googleapis/gapic-generator/pull/1984


garrettjonesgoogle:
Labeling this as `do not merge` because I neglected to update grpc metadata generation.

garrettjonesgoogle:
@lukesneeringer @michaelbausor I have updated this to use the new package metadata format for grpc metadata generation and updated relevant tests - please take another look.

artman/tasks/gapic_tasks.py
```diff
@@ -126,7 +126,7 @@ def execute(self, language, toolkit_path, descriptor_set, service_yaml,
                         for yaml in service_yaml]
         args = [
             '--descriptor_set=' + os.path.abspath(descriptor_set),
-            '--package_yaml=' + os.path.abspath(package_metadata_yaml),
+            '--package_yaml2=' + os.path.abspath(package_metadata_yaml),
```
michaelbausor: How does this affect running artman on existing APIs? Is there any googleapis update required?

garrettjonesgoogle: This file is generated purely within the scope of executing artman - you can almost view the file as an object that is being passed from artman to toolkit, except that it is being passed through a file. Nothing needs to change in googleapis.

PR #425, opened 2018-05-30
Removing unused batch stuff

garrettjonesgoogle:


lukesneeringer:
I am actually delegating this to @michaelbausor, as I do not know if it is used or not.

garrettjonesgoogle:
My read of https://github.com/googleapis/artman/blob/master/artman/cli/main.py#L328 is that only the pipelines listed in the code there are the ones that can be executed.

lukesneeringer:
I believe that to be accurate, but would still feel more comfortable with Michael's sanity check. If he does not respond promptly, I will give you a green checkmark.

garrettjonesgoogle:
I am sad too :-( Oh well, our end-state will be better anyway with Bazel.

PR #430, opened 2018-06-01
Bump to version 0.11.0

garrettjonesgoogle:


michaelbausor:
Can we bump the toolkit version?

garrettjonesgoogle:
I had already bumped it in the prior release, but not the latest latest version, so I went ahead and grabbed the latest commit as of now.

PR #437, opened 2018-06-19
Add a --version flag.

lukesneeringer:
This commit adds a  flag, for salient reasons. (I got tired of having this be an error.)

@garrettjonesgoogle or anyone else, please merge outright upon approval (no need to wait for me to do it).


lukesneeringer:
Note: No unit tests are necessary because the CLI methods are already tested, and this functionality is part of argparse.

PR #439, opened 2018-06-21
Converting gapic_api_yaml/service_yaml to strings internally

garrettjonesgoogle:
- Before: `gapic_yaml` & `service_yaml` were each added to a single-item list (as `gapic_api_yaml` & `service_yaml`)
- After: the original `gapic_yaml` & `service_yaml` values are plumbed through everywhere

This conversion to a list was done because we used to merge with the legacy language configs that have since been deprecated. 

Net deletion!

### DIFF
--- a/test/tasks/test_gapic.py
+++ b/test/tasks/test_gapic.py
@@ -44,7 +44,7 @@ class GapicConfigGenTaskTests(unittest.TestCase):
             descriptor_set='/path/to/descriptor_set',
             organization_name='google-cloud',
             output_dir='/path/to/output',
-            service_yaml=['/path/to/service.yaml'],
+            service_yaml='/path/to/service.yaml',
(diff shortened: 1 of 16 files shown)


test/tasks/test_gapic.py
```diff
@@ -107,7 +107,7 @@ class GapicConfigMoveTaskTests(unittest.TestCase):
     @mock.patch.object(gapic_tasks.GapicConfigMoveTask, 'exec_command')
     def test_execute(self, exec_command):
         task = gapic_tasks.GapicConfigMoveTask()
-        task.execute('/path/src', ['/path/dest'])
+        task.execute('/path/src', '/path/dest')
```
lukesneeringer: Did the signature of `GapicConfigMoveTask()` change, or was this just wrong before?

PR #441, opened 2018-06-25
Update to latest protobuf

jbolinger:
Updating protobuf to latest 3.6.0 release.


jbolinger:
Tests are now all passing. PTAL.

Dockerfile
```diff
@@ -33,22 +33,22 @@ RUN apt-get update \
     python3-pip \
   && rm -rf /var/lib/apt/lists/*
 
-# Install protoc 3.5.1.
+# Install protoc 3.6.0.
 # Currently, the stable version is still 2.6.x, which can only handle proto2
```
garrettjonesgoogle: Is this old comment useful or relevant?

jbolinger: I suspect it's not (certainly wasn't to me)

jbolinger:
@garrettjonesgoogle is there a simple way to do that? 

Perhaps more importantly, shouldn't the CI be doing that anyway? I haven't looked under the hood but I thought that's what the smoke tests were for. Is that what the task that is currently being skipped does?

garrettjonesgoogle:
I don't know of a simple way to do that. The point of CI for artman is to smoke test the generation process, not the library calling process; that's a separate type of smoke test that I don't think is automatically called anywhere currently.

jbolinger:
@garrettjonesgoogle if there is not a simple way to do this can we instead have language owners review the changelog for the [most recent release](https://github.com/google/protobuf/releases) and merge if there aren't any objections? 

Unless there are major issues we need to get on this because it's blocking the release of new features for us as well.

garrettjonesgoogle:
Ok let's go with that.

Dockerfile
garrettjonesgoogle: Maybe remove it now?

jbolinger:
@alexander-fenster @pongad @lukesneeringer @chrisdunelm (or @jskeet) could you please review the changes in protobuf 3.6.0 [here](https://github.com/google/protobuf/releases) and +1 or approve this PR if you're okay to upgrade the tools to use this release. 

Ruby, PHP, & Java are good to go.

pongad:
Unless there's a breaking change to protobuf wire format itself, Go shouldn't be affected. LGTM

jskeet:
As far as I'm aware it should be fine to do this for C#.

alexander-fenster:
I'm fine since Node.js does not use `protoc` now.

PR #442, opened 2018-06-27
Update toolkit to latest

andreamlin:


andreamlin:
PTAL

andreamlin:
Hmm I should fix the golden/smoke tests before checking in.

andreamlin:
Ahh looks like the Dockerfile has already been updated.

PR #443, opened 2018-06-27
Always use latest version of toolkit

andreamlin:
Require that artman Docker image use the latest version of toolkit.


alexander-fenster:
It has obvious pros and cons. I tend to disagree with the idea since it makes it very difficult to rollback to the latest known version in case if gapic-generator is broken in master.

andreamlin:
PTAL
I'd want a +2 approval to check this in.

garrettjonesgoogle:
This means builds aren't repeatable, and it also means we can't tell if we're up to date or not. To me, the task description "Make Artman use the latest toolkit" means that it should watch for new commits to gapic-generator, and roll them through an automatic deployment process.

andreamlin:
Yeah I'm going to think about this more.

@alexander-fenster if gapic-generator is broken in master, then gapic-generator should be rolled back.

@garrettjonesgoogle Yeah the main problem is that each Docker image should be the same always.

Closing. nvm :)

alexander-fenster:
I'm working on artman auto-release based on GitHub releases which will make the process _much_ easier. ETA this week.

PR #449, opened 2018-07-09
chore: auto-release

alexander-fenster:
This is the first step of automating release process in Artman. Whenever a release is tagged with `v[\d.]+` (e.g. `v0.13.1`) and tests pass, the code will be released to Docker and PyPI automatically.

I expect the release code to fail as I haven't had a chance to test it in a real CircleCI run, only locally, so expect some fix PRs after this one is merged.

Note: smoke tests are now failing. I'm working on fixing them in a different branch and will merge this PR when smoke tests are fixed.


alexander-fenster:
@lukesneeringer @garrettjonesgoogle @andreamlin Ping on this please. I want to use it for the next release :)

andreamlin:
This is great!

I'm not terribly familiar with circle config, so I'll let @garrettjonesgoogle or @lukesneeringer give approval.

.circleci/config.yml
```diff
@@ -98,6 +98,42 @@ jobs:
...
+      - run:
+          name: Decrypt PyPI configuration file
+          command: |
+            openssl aes-256-cbc -d -in .circleci/.pypirc.enc -out ~/.pypirc -k "${PYPIRC_ENCRYPTION_KEY}"
+      - run:
+          name: Push Artman Docker tagged image
+          command: |
+            if [ "${DOCKER_EMAIL}" == 'googleapis-publisher@google.com' ]; then
+              # extract version number from setup.py
+              ARTMAN_VERSION=`awk -F"'" '/version=/ { print $2; }' setup.py`
```
garrettjonesgoogle: Do we perhaps want to add a note into setup.py that this version line is scraped for this purpose?

.circleci/config.yml
```diff
@@ -98,6 +98,42 @@ jobs:
...
+              # release to PyPI
+              echo "Releasing artman version ${ARTMAN_VERSION} to PyPI."
+              python3 setup.py sdist upload
+              # push to docker
+              docker login -e "${DOCKER_EMAIL}" -u "${DOCKER_USER}" -p "${DOCKER_PASS}"
+              docker tag "${IMAGE_HASH}" "googleapis/artman:${ARTMAN_VERSION}"
```
garrettjonesgoogle: How does `${IMAGE_HASH}` get set?

.circleci/config.yml
```diff
@@ -98,6 +98,42 @@ jobs:
...
+              docker push "googleapis/artman:${ARTMAN_VERSION}"
```
garrettjonesgoogle: Don't we need to tag & push `googleapis/artman:latest`?

.circleci/config.yml
```diff
@@ -186,3 +220,11 @@ workflows:
           filters:
             branches:
               only: master
+      - release:
+          requires:
+            - build_and_release_docker_image
+          filters:
+            branches:
+              ignore: /.*/
```
garrettjonesgoogle: What exactly does this match?

.circleci/config.yml
alexander-fenster: I actually have an idea for some refactoring of `setup.py` and `Dockerfile` related to versions and commit hashes, will update this PR with that.

.circleci/config.yml
alexander-fenster: My bad. Should tag by name here. Fixing

.circleci/config.yml
alexander-fenster: We do it in another step but I agree it's better to do it here with the same image we just built. Fixing

.circleci/config.yml
alexander-fenster: Does not match any commits to master, matches tags only. We use it for Node packages auto-release.

https://github.com/googleapis/nodejs-pubsub/blob/c400db220d5fcf7bbcbf55927879d2c275b00132/.circleci/config.yml#L52

alexander-fenster:
v2: a little bit of refactoring here.

- All parameters in `Dockerfile` moved to the top as environment variables, much easier to change
- PR feedback from @garrettjonesgoogle 
- A quick'n'dirty script to fetch and update latest commits and versions to prevent any manual edits to `Dockerfile` and `setup.py`. I'm not a Python person at all, so all Python feedback is greatly appreciated.

release.py
```diff
@@ -0,0 +1,139 @@
...
+        usage()
+        return 1
+
+    abspath = os.path.abspath(argv[0])
+    root_dir = os.path.dirname(abspath)
+    dockerfile = os.path.join(root_dir, 'Dockerfile')
+    setup_py = os.path.join(root_dir, 'setup.py')
+
+    sha = None
+    if len(argv) > 2:
```
garrettjonesgoogle: Consider using argparser for command-line args - take a look at the Python version bumping script in google-cloud-java: https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/utilities/bump_versions.py#L117

release.py
```diff
@@ -0,0 +1,139 @@
...
+# limitations under the License.
+"""Release new version of artman.
+
+This script updates SHA hashes of latest commits for both gapic-generator
+and googleapis, and updates version number for subsequent (auto-)release.
+
+Usage:
+$ python3 release.py googleapis [sha]      # update googleapis commit hash
+$ python3 release.py gapic-generator [sha] # update gapic-generator commit hash
+$ python3 release.py 1.12.4                # update artman version
```
garrettjonesgoogle: Since this script doesn't actually release anything, I'd prefer a different name, like maybe `update_versions.py` or `bump_versions.py` or whatever.

Also, the need to call it multiple times feels a little odd. I would prefer something like the following:

```
# fetch automatically
$ python3 update_versions.py --fetch-googleapis --fetch-gapic-generator 1.12.4`

# manual shas
$ python3 update_versions.py --update-googleapis <sha> --update-gapic-generator <sha> 1.12.4`
```

release.py
```diff
@@ -0,0 +1,139 @@
...
+    try:
+        yaml = YAML()
+        with open(artman_config) as config_yaml:
+            config = yaml.load(config_yaml)
+            github_username = config['github']['username']
+            github_token = config['github']['token']
+            auth = requests.auth.HTTPBasicAuth(github_username, github_token)
+    except:
+        print('Cannot get GitHub token from %s, trying anonymously...' %
+              artman_config)
```
garrettjonesgoogle: Why can't we always do it anonymously?

release.py
alexander-fenster: I spent some time thinking about it. Basically, updating gapic-generator commit sha does not force us to release a new artman version immediately, and vice versa, so if we go with named parameters, we'll need one for version as well.  How about this:

```sh
$ python3 update_versions.py --fetch-googleapis
$ python3 update_versions.py --set-googleapis <sha>
$ python3 update_versions.py --fetch-gapic-generator
$ python3 update_versions.py --set-gapic-generator <sha>
$ python3 update_versions.py --set-version <version>
```
(options can be combined in one invocation)

Does it make sense?

garrettjonesgoogle: My thought was that if someone wanted to only update a sha, they could just modify the file without having to go through the complexity of running the command - the value I see for this is the "maintenance" case of "just update everything & prepare for the next release".

release.py
alexander-fenster: Because of the rate limit. When running on gLinux machine, it's very likely you'll get 403 from GitHub API because someone has eaten 60 queries per hour per IP that are allowed for anonymous queries.

```sh
$ curl 2>/dev/null https://api.github.com/rate_limit | jq .resources.core
{
  "limit": 60,
  "remaining": 31,
  "reset": 1531350123
}
```

Authenticated users get higher limits. It's not that important so the script is just doing its best trying to locate a config in the default place and does not fail if there is no config, or no token there.

alexander-fenster:
Updated based on your feedback - please take a look.

update-versions.py
```diff
@@ -0,0 +1,184 @@
...
+    with open(filename) as f:
+        for line in f:
+            line = re.sub(regex, r'\g<1>' + version + r'\g<2>', line)
+            lines.append(line)
+    with open(filename, 'w') as f:
+        f.write(''.join(lines))
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Prepare new artman release.')
```
garrettjonesgoogle: It looks like half of this file is argument validation, which feels a little over-engineered to me to just update a couple sha values. I'd really like to treat this as a "script" - just have it do some straightforward actions. I'd propose removing the `--set-X` functionality since we hardly ever need that, and it's trivial to just modify the hashes manually when we need to.

update-versions.py
```diff
@@ -0,0 +1,184 @@
+# Copyright 2018 Google LLC
```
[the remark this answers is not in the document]
alexander-fenster: Sounds good - will do. I'll leave the `Dockerfile` and `setup.py` changes though since they are related to `.circleci/config.yml` logic to extract the version.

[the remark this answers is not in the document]
garrettjonesgoogle: Yes, I agree that it makes sense to keep the change which extracts the versions up.

alexander-fenster:
Moving the script to a separate PR: #463.

PR #451, opened 2018-07-10
fixing corner cases of config refactoring

alexander-fenster:
This PR fixes two corner cases that were causing smoke tests failures (mostly with `artman_base.yaml`) after recent config refactoring. The two corner cases that this PR fixes are:
- empty `proto_deps`
- no `service_yaml`

PR #453, opened 2018-07-10
Removing remote functionality

garrettjonesgoogle:


alexander-fenster:
Let's have #451 merged first as it fixes smoke tests, after that we'll be able to tell if anything is broken by this change (hopefully not).

PR #454, opened 2018-07-10
Allowing artifact definitions to be omitted

garrettjonesgoogle:


artman/config/loader.py
```diff
@@ -37,22 +37,42 @@
 INVALID_USER_CONFIG_ERROR_MESSAGE_FORMAT = 'Artman user YAML %s is invalid.'
 
 
+ARTIFACT_MAPPING = {
```
lukesneeringer: If artman survives, I do not think this is ultimately a good approach, because it embeds language logic in the code in a non-extensible way.

That said, I _do_ think it is fine for now.

garrettjonesgoogle: Agreed - I don't really like the approach of fixed strings containing both the language and the artifact; those should be separate parameters which are passed through. But, since we're trying to maintain surface stability, this should probably be retained.

PR #456, opened 2018-07-10
run smoke tests only in master branch

alexander-fenster:
We still run artman for some APIs (speech, pubsub, etc.) before this step so let's run the complete smoke testing, which takes about 40 minutes, only for master commits.


alexander-fenster:
@lukesneeringer are you ok with this idea?

alexander-fenster:
Since no one objects I consider this approved.

PR #457, opened 2018-07-11
Use absolute path for discovery doc

andreamlin:
This allows the artman command line call to run from any directory, e.g. google-cloud-java, not just from inside the discovery_artifact_manager directory.


andreamlin:
PTAL

test/tasks/test_gapic.py
```diff
@@ -189,7 +189,8 @@ def test_execute(self, exec_command, check_output):
             organization_name='google-cloud',
             package_metadata_yaml='/path/to/pmy.yaml',
             discovery_doc="/path/to/compute.v1.json",
-            toolkit_path='/path/to/toolkit'
+            toolkit_path='/path/to/toolkit',
+            root_dir='root_dir'
```
garrettjonesgoogle: 1. A root dir should start with a slash
2. This doesn't seem to be influencing the commands below
3. This usage of an absolute path doesn't seem necessary for gapic generation; I think it's because I invoke gapic generation using the full path to the gapic config, so it's already an absolute path.

andreamlin: 1, 3: What if I want to run artman from a path relative to my current directory?
2: Fixed the test.

garrettjonesgoogle: 1, 3: That also works too (currently for GAPIC)

andreamlin: The problem is that artman resolves all filepaths in the config to be relative to the config filepath and assumes that the artman config file is at the highest (or equal) level directory compared to the files in the config. 
For example, in googleapis, for `google/pubsub/artman_pubsub.yaml`, all the pubsub API files are located at or below that directory level, like `google/pubsub/v1/pubsub.proto`.

However, in discovery-artifact-manager, the relevant discovery document is not in the same directory path as the artman config.
For example, for `gapic/google/compute/artman_compute.yaml`, the relevant discovery doc is at `discoveries/compute.v1.json`.

So there are two options:
1. this proposed PR, which allows specifying the base `discovery-artifact-manager` location
2. Removing the check in https://github.com/googleapis/artman/blob/master/artman/config/loader.py#L198, which would allow specifying the discovery_doc path as "../../discoveries/compute.v1.json" relative to the artman config filepath.

artman/tasks/gapic_tasks.py
```diff
@@ -128,14 +128,13 @@ class DiscoGapicCodeGenTask(task_base.TaskBase):
 
     def execute(self, language, toolkit_path, discovery_doc,
         gapic_yaml, package_metadata_yaml,
-        gapic_code_dir, api_name, api_version, organization_name):
+        gapic_code_dir, api_name, api_version, organization_name, root_dir):
```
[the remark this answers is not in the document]
garrettjonesgoogle: Taskflow populates this for us out of a global dictionary. I don't think this parameter list is a problem.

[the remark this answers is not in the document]
alexander-fenster: (after some thoughts) well, it's probably OK

alexander-fenster:
I might be missing something - it's unclear for me how the root dir will be used. Why it only applies to one parameter (discovery doc)?

andreamlin:
@alexander-fenster 
(Copied from a previous comment)
The problem is that artman resolves all filepaths in the config to be relative to the config filepath and assumes that the artman config file is at the highest (or equal) level directory compared to the files in the config.
For example, in googleapis, for google/pubsub/artman_pubsub.yaml, all the pubsub API files are located at or below that directory level, like google/pubsub/v1/pubsub.proto.

However, in discovery-artifact-manager, the relevant discovery document is not in the same directory path as the artman config.
For example, for gapic/google/compute/artman_compute.yaml, the relevant discovery doc is at discoveries/compute.v1.json.

PTAL

PR #460, opened 2018-07-11
Update toolkit fatJar comments

andreamlin:
Document that one should run `./gradlew clean fatJar` in the toolkit repo to refresh the toolkit code that artman calls on.


andreamlin:
PTAL

artman/config/proto/user_config_pb2.py
```diff
@@ -38,14 +38,14 @@
       has_default_value=False, default_value=None,
       message_type=None, enum_type=None, containing_type=None,
       is_extension=False, extension_scope=None,
-      options=None),
+      options=None, file=DESCRIPTOR),
```
garrettjonesgoogle: Why is this changing in this PR?

andreamlin: I ran the protoc script and it happened!
This also happens after running the protoc script on the master version of user_config.proto.

The Dockerfile says to use protoc 3.6.0, and after downloading that, I used v3.6.0 to regenerate this file again, which results in a couple more updates.

andreamlin: This does break users who are not on protobuf==3.6.0, but that is also specified in Dockerfile.

garrettjonesgoogle: It kind of makes this PR non-cohesive though - would you consider splitting it into two PRs?

andreamlin: Done!

PR #462, opened 2018-07-11
Recompile user_config.proto with protoc v3.6.0

andreamlin:
Just ran ` protoc --proto_path=. --python_out=. user_config.proto `


andreamlin:
lol that was fast

PR #467, opened 2018-07-13
Make src_proto_paths optional

andreamlin:
`src_proto_paths` is not needed in discogapic gen.


garrettjonesgoogle:
Looks like you have some test failures...

andreamlin:
Fixed!

PTAL

PR #473, opened 2018-07-25
Remove `publish` target from artman and config

andreamlin:
Remove (github and local) publishing artman configs, as these are not needed by ACTools or Yoshi team.

This reduces the size of the artman configs.

The old `artman.tasks.publish.noop` file is renamed as `artman.tasks.emit_success`, which more accurately describes that task.

In the CircleCI smoke tests, change all `publish` targets to `generate`.

Eventually, we would be able to remove the `generate` target because it's the only option, and it would be frivolous to specify it if there are no other options.


andreamlin:
Waiting for smoke tests to pass before sending out for review.

andreamlin:
PTAL

andreamlin:
1. Because googleapis is synced from internal repos, wouldn't the changes get overridden by internal pushes?

2. Sure.

garrettjonesgoogle:
1. Of course - the change would need to be done in the internal files (I thought that was implied...)

andreamlin:
1. Any changes made to the internal files would get overridden as soon as an API producer regenerates or bootstraps a config using a version of artman without this PR.

garrettjonesgoogle:
1. API producers won't regenerate artman files. In the case of config bootstraps, I think we can handle the 0-2 cases that might happen in the "gap" on a one-off basis.

andreamlin:
Ok, made an internal PR for removing all instances of the `publish_targets`.
Also made an internal PR for deprecating g3artman.

garrettjonesgoogle:
I think you mean a PR to deprecate the *publish feature* in g3artman?

andreamlin:
Ok, (1) and (2) are done!
PTAL

artman/config/proto/config.proto
```diff
@@ -266,6 +252,7 @@ message Artifact {
   // [dependencies.yaml](https://github.com/googleapis/googleapis/blob/master/gapic/packaging/dependencies.yaml)
   PackageVersion package_version = 14;
 
+  // DEPRECATED.
```
garrettjonesgoogle: Maybe use `UNSUPPORTED` (here and below)

artman/pipelines/code_generation.py
```diff
@@ -23,6 +23,7 @@
 from artman.pipelines import pipeline_base
 from artman.tasks import io_tasks
 from taskflow.patterns import linear_flow
+from artman.tasks import emit_success
```
garrettjonesgoogle: This doesn't appear to be referenced in this file

artman/pipelines/gapic_generation.py
```diff
@@ -18,14 +18,15 @@
...
 from artman.pipelines import code_generation as code_gen
 from artman.pipelines import grpc_generation as grpc_gen
+from artman.tasks import emit_success
 from artman import tasks
 from artman.utils import task_utils
 
 
 # kwargs required by GAPIC code gen
-_GAPIC_REQUIRED = ['service_yaml', 'gapic_yaml', 'language', 'aspect', 'publish']
+_GAPIC_REQUIRED = ['service_yaml', 'gapic_yaml', 'language', 'aspect',]
```
garrettjonesgoogle: remove the trailing comma

andreamlin: Done.

artman/pipelines/code_generation.py
andreamlin: Done.

artman/config/proto/config.proto
andreamlin: Done.

PR #475, opened 2018-07-25
Creating `aspect` option (ALL, CODE, PACKAGE)

garrettjonesgoogle:
This will only work for C# so far; after https://github.com/googleapis/gapic-generator/pull/2166 , it will work for Java too.


artman/cli/main.py
```diff
@@ -179,6 +179,11 @@ def parse_args(*args):
         type=str,
         help='[Required] Name of the artifact for artman to generate. Must '
         'match an artifact in the artman config yaml.')
+    parser_generate.add_argument(
+        '--aspect',
```
vam-google: Are you sure that aspect is a proper name here? It feels too generic. I knew what this review would be about, but even in this case i was baffled by "aspect" argument. Maybe something more specific like "artifact_type" or something like that?

artman/tasks/gapic_tasks.py
```diff
@@ -116,8 +117,19 @@ def execute(self, language, toolkit_path, descriptor_set, service_yaml,
         if service_yaml:
             args = args + ['--service_yaml=' + os.path.abspath(service_yaml)]
         args = args + gapic_args
+
+        gapic_artifact = ''
+        if aspect == 'ALL':
```
vam-google: It would be nice if `aspect` name (`ALL`) matched `gapic_artifact` (`LEGACY_GAPIC_AND_PACKAGE`). Note, `CODE` and `PACKAGE` already match (`LEGACY_` prefix is fine).

artman/config/loader.py
```diff
@@ -54,7 +54,7 @@
 }
 
 
-def load_artifact_config(artman_config_path, artifact_name):
+def load_artifact_config(artman_config_path, artifact_name, aspect=None):
```
vam-google: Will  `aspect='ALL'` be appropriate here? I hope it can allow to not specify default value for aspect parameters in `test_code_generation.py` file (multiple times).

artman/pipelines/grpc_generation.py
```diff
@@ -25,156 +25,84 @@ class GrpcClientPipeline(code_gen.CodeGenerationPipelineBase):
...
+        self.gen_code = kwargs['aspect'] == 'ALL' or kwargs['aspect'] == 'CODE'
+        self.gen_pkg = kwargs['aspect'] == 'ALL' or kwargs['aspect'] == 'PACKAGE'
 
     def get_tasks(self, **kwargs):
         tasks = self.get_grpc_codegen_tasks(**kwargs)
         tasks += self._get_publish_tasks(**kwargs)
         return task_utils.instantiate_tasks(tasks, kwargs)
 
     def get_grpc_codegen_tasks(self, **kwargs):
+        if self.language == 'java':
```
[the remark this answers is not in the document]
garrettjonesgoogle: There's not really much that is shared, and there's not really much logic either, it's basically just providing a sequence of commands. I feel like trying to share stuff here creates more abstraction complexity than is worth it. I'm trying to cut out abstractions and make this as concrete as possible.

artman/cli/main.py
garrettjonesgoogle: I tried out the `artifact_type` name initially, but then it felt like it was too easy to confuse with `artifact_name`. Any other ideas? Maybe `artifact_aspect`?

artman/tasks/gapic_tasks.py
garrettjonesgoogle: The problem is that `aspect` is also used for just PROTOBUF generation - you can do `PROTOBUF` + `CODE`. (on the command line: `artman ... java_proto --aspect CODE`)

artman/config/loader.py
garrettjonesgoogle: Oh, I could actually remove the aspect parameters from `test_code_generation.py` as it is; `aspect=None` has the same result as `aspect='ALL'`.

garrettjonesgoogle: So it turns out that `test_code_generation.py` needs the parameter to be set, because `test_code_generation.py` doesn't call `load_artifact_config` directly.

artman/pipelines/grpc_generation.py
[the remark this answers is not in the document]
vam-google: Can you at least put them in separate methods (each `if` is a method)? maybe even then you can use map instead of if-else statements (i'm not sure, but worth trying)

artman/cli/main.py
vam-google: Definitelly not `artifact_aspect` it is even worse. How is coexistence of `name` and `type` is confusing? It is a very-very common pair of properties.

artman/tasks/gapic_tasks.py
vam-google: Then maybe `CODE_AND_PACKAGE` instead of `ALL`? There is no way to findout what `ALL` without reading docs/comments.

artman/cli/main.py
vam-google: Definitely not `artifact_aspect` it is even worse. How is coexistence of `name` and `type` confusing? It is a very-very common pair of properties.

artman/pipelines/grpc_generation.py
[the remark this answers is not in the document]
vam-google: Can you at least put them in separate methods (each `if` is a method)? maybe even then you can use map instead of if-else statements (i'm not sure, but worth trying)

artman/cli/main.py
garrettjonesgoogle: It's confusing because `artifact_name` and `artifact_type` where wouldn't be related in the same way as the name of a variable and the type of a variable are related. The existing `artifact_name` is actually a type, and would actually be better named if it was `artifact_type`. The term `aspect` here is meant to invoke the concept of aspect-oriented programming, indicating that it is a cross-cutting concern.

artman/tasks/gapic_tasks.py
vam-google: Then maybe `CODE_AND_PACKAGE` instead of `ALL`? There is no way to findout what `ALL` without reading docs/comments.

artman/pipelines/grpc_generation.py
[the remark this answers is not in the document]
garrettjonesgoogle: Ok I got a map to methods to work.

garrettjonesgoogle:
`ALL` vs `CODE_AND_PACKAGE`: I considered the second, but I wanted to allow for adding a third item, and the default should be to include everything I think.

garrettjonesgoogle:
I split apart the long function, PTAL

Issue #479, opened 2018-07-26
smoke tests are failing


alexander-fenster:
#475 broke smoke tests, three invocations fail:

```
Failure:
Failed to generate java_gapic of /tmp/googleapis_smoketest/google/datastore/admin/artman_datastore_admin.yaml.
Failed to generate java_gapic of /tmp/googleapis_smoketest/google/longrunning/artman_longrunning.yaml.
Failed to generate gapic_config of /tmp/googleapis_smoketest/google/monitoring/artman_monitoring.yaml.
```

All three fail with the following in the log (available as CircleCI build artifact [here](https://7515-55101236-gh.circle-artifacts.com/0/tmp/smoketest.log)):

```
[35martman> [36mCreating GapicOnlyClientPipeline.[0m
[35martman> [31mTraceback (most recent call last):
  File "/artman/artman/cli/main.py", line 67, in main
    **pipeline_kwargs)
  File "/artman/artman/pipelines/pipeline_factory.py", line 53, in make_pipeline
    return cls(**kwargs)
  File "/artman/artman/pipelines/gapic_generation.py", line 80, in __init__
    GapicOnlyTaskFactory(**kwargs),
TypeError: object() takes no parameters
```

This is blocking automated releases. Can I ask you to take a look and fix, or reassign as appropriate?

Thanks!

garrettjonesgoogle:
`java_gapic` in artman_datastore_admin.yaml should not be `GAPIC_ONLY`. I am pushing an internal CL to switch it to `GAPIC`.

garrettjonesgoogle:
https://github.com/googleapis/gapic-generator/pull/2169 in gapic-generator needs to be pulled in to fix the `gapic_config` artifact of artman_monitoring.yaml.

alexander-fenster:
OK so I'll update hashes for gapic-generator and googleapis and it might fix these two. What about `artman_longrunning.yaml`?

garrettjonesgoogle:
That is fixed in https://github.com/googleapis/artman/pull/480

PR #480, opened 2018-07-26
Fixing GAPIC_ONLY

garrettjonesgoogle:
Fixes https://github.com/googleapis/artman/issues/479


garrettjonesgoogle:
(I verified the fix locally btw)

PR #485, opened 2018-08-07
Enable grpc package gen for ruby and python

andreamlin:
To generate common proto packages for https://github.com/googleapis/gapic-generator/issues/2204 and https://github.com/googleapis/gapic-generator/issues/2209.


andreamlin:
PTAL

landrito:
Generating the common protos gems through artman will not work for ruby since there needs to be two specific gems, one that only provides the types, and another that depends on the former as well as adds the operations service and depends on grpc. 

These two packages were made so that grpc can depend on these types without introducing a circular dependency.

garrettjonesgoogle:
@landrito Java has the same split between common protos and the operations grpc stub.

landrito:
How are you using artman to generate the types without the operations stub. I only see the grpc target in the artman_core.yaml.

PR #490, opened 2018-08-21
Add api-common-protos repo to Docker image

andreamlin:
googleapis/api-common-protos should be the source of truth for the API common protos. Include a version source repo for api-common-protos in the artman Docker image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -173,7 +173,7 @@ RUN git clone https://github.com/googleapis/googleapis \
   && rm -rf /googleapis/.git/
 RUN git clone https://github.com/googleapis/api-common-protos \
   && cd api-common-protos \
-  && git checkout COMMON_PROTOS_HASH \
+  && git checkout $COMMON_PROTOS_HASH \
   && cd .. \
   && rm -rf /api-common-protos/.git/
 RUN git clone https://github.com/googleapis/gapic-generator toolkit \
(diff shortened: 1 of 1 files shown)


garrettjonesgoogle:
@lukesneeringer is this really the source of truth yet?

lukesneeringer:
Artman currently still uses what is in googleapis/googleapis. I think it would be possible and worthwhile to change this (as this PR does), and effectively make api-common-protos the source of truth now.

Thinking through the benefits and risks:

  * **Benefit:** This is where we want to end up. :smile:
  * **Benefit:** Artman no longer needs to keep googleapis inside its Docker image at all (note: this PR should remove it). That also should make it stop writing over local googleapis checkouts in-place.
  * **Benefit/Risk:** No automatic sync is available from `third_party` in internal version control to googleapis/api-common-protos. This is arguably a good thing since it is often not sufficient to drop a common proto there and just assume everything will magically work (it rarely does).
  * **Benefit/Risk:** No mechanism to have common protos in private.
  * **Risk:** Will any other workflows break when we remove the common protos from googleapis?

andreamlin:
"no automatic sync is available from third_party in internal version control to googleapis/api-common-protos": Should we set up the CI build for googleapis/googleapis to pull from api-common-protos, and then set up an internal sync for api-common-protos? My opinion is that it's better to know ASAP when a regression is introduced, which helps with debugging and reverting.

andreamlin:
We can wait to discuss this PR after the Tuesday morning meeting.

alexander-fenster:
Year old PR, closing.

PR #491, opened 2018-08-22
Prepare for 0.15.2 release

vchudnov-g:


vchudnov-g:
Is there anything else I need to do to make this a proper release of artman that will not break anything?

I'm interested in this so I can have the latest toolkit changes in artmanflow. But I don't want to break any other tooling, obviously.

andreamlin:
See go/artman-release-instructions for the rest of artman release
instructions.

The artman circleCI should be ok; it just generates all possible targets
from googleapis. If you haven't changed the surface of gapic-generator,
then I don't see where artman would break...

vchudnov-g:
PTAL @jbolinger @andreamlin

jbolinger:
LGTM @vchudnov-g

PR #496, opened 2018-09-12
Prepare for 0.15.3 release

jbolinger:
Prep for a minor release needed by the Yoshi team.


andreamlin:
Hmm can you triage the smoke test failures?

jbolinger:
It looks like it has something to do with this `experimental_features` field (maybe it was removed?):
https://github.com/googleapis/artman/blob/fbde0386d7ebc697410ddf3c431549df10b00682/test/tasks/data/googleapis/google/pubsub/v1/pubsub_gapic.yaml#L79

@pongad is this related to what you were doing? I recall seeing some changes related to iam?

andreamlin:
Do the failing artman commands pass if you ran them locally?

jbolinger:
nope, failing locally too. I think it's because I didn't update the googleapis hash. That `experimental_features` flag seems to no longer be there. We'll see in a moment...

jbolinger:
@andreamlin I think this is good to go now. Can you take another look?

PR #498, opened 2018-09-18
update whitelist documentation

pongad:


test/smoketest_artman.py
```diff
@@ -98,8 +98,8 @@ def run_smoke_test(root_dir, log):
     logger.info('Success:')
     for msg in success:
         logger.info(msg)
-    logger.info('Warning:')
-    for msg in warning:
+    logger.info('Known Failures:')
```
andreamlin: Can we call these "Known Whitelisted Failures"?

test/smoketest_artman.py
```diff
@@ -98,8 +98,8 @@ def run_smoke_test(root_dir, log):
...
+    for msg in known_fail:
         logger.info(msg)
     logger.info('Failure:')
     for msg in failure:
```
andreamlin: While we're at it, if there are no failures, can we either print "none" under "Failure:" or not print "Failure:" at all?
Otherwise we get something like 
```
Failure:
bash exited 0. unknown command
```
and people might think that bash failed somehow

pongad: The suggestion sounds good, but why do we get the `unknown command` line?

PR #499, opened 2018-09-19
Apply pandoc to descriptor set

michaelbausor:
Attempts to fix https://github.com/googleapis/gapic-generator/issues/1587 by applying pandoc to the descriptor set. Looking for validation on the general approach before adding tests.


michaelbausor:
@lukesneeringer can you PTAL to confirm that this general approach is reasonable?

PR #504, opened 2018-09-26
extract version from Dockerfile

alexander-fenster:
I'm a little bit tired of changing version number in two files and it's error prone. Let's store the version only in `Dockerfile`, and make `setup.py` parse it from there. Adding a little bit more logic into `setup.py` might not be a good idea, but having just one version number across the code looks like a good idea to me.


setup.py
```diff
@@ -19,13 +19,23 @@
...
 
-# The following line is parsed by CI scripts (see .circleci/config.yml)
-# and by release.py. Please keep the format.
-current_version = '0.15.5'
-
 cur_dir = os.path.realpath(os.path.dirname(__file__))
+
+# version is defined in Dockerfile: ENV ARTMAN_VERSION version
+current_version = None
+with io.open('%s/Dockerfile' % cur_dir) as dockerfile:
```
pongad: is it worth detecting path separator from environment?

setup.py
```diff
@@ -19,13 +19,23 @@
...
+    for line in dockerfile:
+        match = re.match('^ENV ARTMAN_VERSION (\S+)$', line)
+        if match:
+            current_version = match.group(1)
```
pongad: Do we want to `break`?

setup.py
alexander-fenster: Yes! changed here and below

setup.py
alexander-fenster: We can (in hope that no one will define version twice :) )

alexander-fenster:
@lukesneeringer what do you think?

PR #517, opened 2018-10-16
whitelist resultstore

hzyi-google:


test/smoketest_artman.py
```diff
@@ -62,6 +62,11 @@
     # java.lang.IllegalStateException: A HttpRule option must be defined.
     # (affects only PHP that generates code for JSON over HTTP)
     'php_gapic@google/cloud/bigquery/storage/artman_bigquerystorage_v1beta1.yaml',
+
+    # whitelisted because the gapic config for interface ResultStoreUpload does not exist
+    # already let the team know
+    # TODO(hzyi): remove it from whitelist once API owners fix it
+    'gapic@google/devtools/resultstore/artman_resultstore.yaml',
```
alexander-fenster: I'm not sure it will work since it's a regex and we use `re.match` for checking, which looks at the beginning of the line.

https://github.com/googleapis/artman/blob/20839f4c23335969f21f865230a10a5584955243/test/smoketest_artman.py#L151
```suggestion
    '.+@google/devtools/resultstore/artman_resultstore.yaml',
```

hzyi-google:
@alexander-fenster It seems the api owner have updated //third_party/googleapis/. Is there a way to force sync googleapis/googleapis and kick off a CI? We won't need to whitelist it if the problem is fixed.

alexander-fenster:
It works!

image

Thanks for working on it @hzyi-google! I guess we can close this PR now.

hzyi-google:
Thanks for the help Alex.

Issue #478, opened 2018-07-26
remove gapic_config@artman_bigtableadmin.yaml from smoketest whitelist


andreamlin:
https://github.com/googleapis/gapic-generator/pull/2169
fixes that artman call, so we can remove it from the white list in smoketest_artman.py

PR #525, opened 2018-11-15
Remove working artman calls from whitelist of smoketest

andreamlin:
Several artman calls have been fixed, and we can remove them from the whitelist.

See the latest smoke test run at https://circleci.com/gh/googleapis/artman/8385 for the list of artman calls that still fail and need to remain in the whitelist.

Cleans up https://github.com/googleapis/artman/issues/478.


andreamlin:
`artman --config google/iam/credentials/artman_iamcredentials_v1.yaml --local generate python_gapic` works for me locally.

andreamlin:
PTAL

PR #528, opened 2018-11-19
Revert version to 0.16.0

andreamlin:
The release didn't go through. Let's revert the version back before the bump. Calling artman locally is wonky when the local version of artman is 0.16.1 but no docker image for artman 0.16.0 exists.


andreamlin:
PTAL

andreamlin:
https://circleci.com/gh/googleapis/artman/8443 fails but this is expected because the fix has not yet been checked in.

PR #529, opened 2018-11-19
Pass in user_config to artman's smoketest

andreamlin:
By parameterizing user_config option, gapic-generator can now use the same smoketest.py test script to run the smoke test.


test/smoketest_artman.py
```diff
@@ -122,14 +123,15 @@ def _parse(artman_yaml_path):
     return config_pb
 
 
-def _generate_artifact(artman_config, artifact_name, root_dir, log_file):
+def _generate_artifact(artman_config, artifact_name, root_dir, user_config, log_file):
```
alexander-fenster: I don't want to be too picky about the test code but here the parameter order is weird, and 5 parameters is a lot :)
We can leave it as is for now (I'd move `user_config` closer to `artman_config` though), but if we ever get more parameters, we'll need to group them into objects. First link in Google: https://refactoring.guru/smells/long-parameter-list

test/golden/artman_golden_test.py
```diff
@@ -81,6 +81,7 @@ def generate_gapic_library(golden_dir, googleapis_dir, output_dir,
         '--local',
         '--root-dir', googleapis_dir,
         '--output-dir', output_dir,
+        '--user-config', "/artman/artman-user-config-in-docker.yaml",
```
alexander-fenster: How will this file appear in the docker image?

test/smoketest_artman.py
andreamlin: Ok, moved the `user_config` parameter closer to the `artman_config`.

Would passing the artman_config as a named arguments be better?

test/golden/artman_golden_test.py
andreamlin: In the docker image, it would be at [/root/.artman/config.yaml](https://github.com/googleapis/artman/blob/e03351421771447e731771c04acb4aa3867c9a39/Dockerfile#L175)

test/smoketest_artman.py
alexander-fenster: I think names arguments would make code more readable. But again, since it's just a test code, we can leave it as is :)

PR #531, opened 2018-11-20
Remove update_versions.py script

andreamlin:
We should only be updating the artman Dockerfile version during releases.

~googleapis is pretty much obsolete; that can probably be safely removed.~
gapic-generator hash can be updated manually, and that can be done on an ad-hoc basis.


andreamlin:
PTAL

@alexander-fenster do you think we still need this update script? At the very least, we should separate out the dependency hash's update from the artman version update.

alexander-fenster:
Yeah, it's probably pretty much useless now.

alexander-fenster:
I just noticed that the contents of this PR and its description don't really match :)

andreamlin:
lol whoops

andreamlin:
PTAL fixed diff

PR #532, opened 2018-11-20
Update version to 0.16.2

andreamlin:
Bumping to 0.16.0. There was a snafu with version 0.16.1, so let's just skip that.

The most recent smoke test passed (https://circleci.com/gh/googleapis/artman/8503)! So this should be good to go.


andreamlin:
PTAL

PR #535, opened 2018-11-30
[Python] Convert docs from commonmark, not md.

lukesneeringer:
This changes the source documentation type from generic Markdown to CommonMark, which seems to be slightly more precise.

This change _may_ address googleapis/gapic-generator#2191 (still testing that), but we should do it regardless of whether it does or not.

Note to reviewers: Please merge on acceptance.


lukesneeringer:
Confirmed that it does fix the problem @crwilcox found in dataproc.

lukesneeringer:
@michaelbausor I could use advice from you on the linting errors. I ran the job with SSH and investigated them, and they look like the RST conversion doing its job. In particular, it is taking the famous troublesome transactions comment in Spanner and rendering it in sections (which Sphinx dislikes, but that's the fault of the underlying proto).

Did you have some post-processing that this change worked around?

michaelbausor:
I don't think so. I'm looking now, but I can't see why this would break.

michaelbausor:
So, when the input type is 'md', the pandoc conversion leaves the headings as:
`# Transactions`
but when it is 'commonmark' it changes them to:
`Transactions`
`============`
I don't know (a) why, or (b) which is correct, but the restructuredtext_lint module doesn't like the second format.

noahdietz:
Hi all, this PR is pretty out of date, but it seems to be the fix for an [issue out of SLO ](https://github.com/googleapis/gapic-generator/issues/2191). Any update?

PR #539, opened 2018-12-13
go: split file across go_package option, not by directory, again

pongad:
Redo of #536


artman/utils/protoc_utils.py
```diff
@@ -198,21 +198,36 @@ def proto_compiler_command(self):
...
-        dirname = os.path.dirname(proto)
-        if dirname not in dirs:
-            dirs[dirname] = [proto]
-        else:
-            dirs[dirname].append(proto)
-    return dirs
+
+    def go_pkg(file):
+        prefix = 'option go_package ='
+        with open(file, encoding='utf-8') as f:
```
andreamlin: Is this line the only line that's changed from the first PR?

pongad: Yup!

andreamlin:
(I re-ran the golden_and_smoke_tests for kicks, but the previous invocation passed.)

PR #545, opened 2019-01-08
Update protobuf dependencies to 1.17.1

jbolinger:
Protobuf dependencies prior to 3.6.0 are causing internal build failures for python and ruby. 

Ruby fails due to the missing `ruby_namespace` option that was added. I'm not sure why Python fails. I've upgraded the dependency across the board. C# and PHP do not currently cause problems but I've updated them anyway. @michaelbausor and @chrisdunelm do you think this will cause any problems?


michaelbausor:
I don't anticipate problems in php

PR #546, opened 2019-01-08
prepare for 0.16.5 release

jbolinger:
Probably should have combined this with the last one...


alexander-fenster:
You did it right, we (@andreamlin and I) found that combining brings a lot of extra problems if CI fails for whatever reason. It's always better to leave the version bumping PR separate (even though it takes more time to wait for the CI).

LGTM after https://circleci.com/gh/googleapis/artman/8798 passes :)

PR #549, opened 2019-01-14
Update gapic-generator and googleapis hash

andreamlin:


andreamlin:
PTAL

test/golden/library_example_gapic.yaml
```diff
@@ -50,7 +50,6 @@ interfaces:
         - shelf
     required_fields:
     - shelf
-    request_object_method: false
```
andreamlin: This config value has been removed from gapic-generator's GAPIC config schema.

PR #567, opened 2019-02-13
release 0.16.13

alexander-fenster:
Let's keep incrementing it up to 0.16.100! That milestone is not too far away.


hzyi-google:
LGTM. Should I just release 0.16.13 for g3artman after Piper is up?

PR #575, opened 2019-03-11
Fix videointelligence bug

andreamlin:
Remove the videointelligence target from the whitelist because the latest gapic-generator fixes that.

Tested locally by running 
```
 artman \
--config google/cloud/videointelligence/artman_videointelligence_v1p3beta1.yaml \
--local generate java_gapic
```


andreamlin:
PTAL

PR #589, opened 2019-03-28
Bump protoc version to 3.7.1

michaelbausor:
This will mean PHP generation picks up the latest changes in protoc, including enum<->string convertion.


andreamlin:
@hzyi-google Was PHP the only blocker for this?

hzyi-google:
I think @jbolinger mentioned protoc 3.7.0 had a few issues([this one](https://github.com/protocolbuffers/protobuf/pull/5861) and [this one](https://github.com/protocolbuffers/protobuf/pull/5867)) in Ruby, but it seems both of them are fixed.

PR #594, opened 2019-04-03
Install vim in artman image

andreamlin:
Vim is really nice for debugging once I'm inside the image, both for CircleCI and in my local artman docker images. It adds 35MB on disk, according to apt-get.


andreamlin:
PTAL

PR #596, opened 2019-04-03
Use v0.0.13 protoc-java-resource-names-plugin

andreamlin:
That dep just got an implementation to process regular resource names as fixed resource names if applicable.

This will allow us to fully remove `fixed_name_resource_whatever` from pubsub_gapic.yaml and gapic-generator.


andreamlin:
Tested with a locally-built artman docker image.

andreamlin:
PTAL

PR #597, opened 2019-04-03
Update gapic-generator hash

andreamlin:
Includes some changes to resource name parsing. 

Goal: Remove explicitly-configured fixed resource names from all GAPIC configs, so we can simplify gapic-generator.


andreamlin:
PTAL

PR #599, opened 2019-04-04
Whitelist bigquery v2

andreamlin:
https://github.com/googleapis/gapic-generator/issues/2691 causes php for bigquery to fail. this seems to be a known issue, since there is another php whitelisted API.

this blocks the artman release


andreamlin:
PTAL

PR #601, opened 2019-04-11
[chore] Update to new Java resource names invocation.

lukesneeringer:
This uses the new invocation mechanism for the Java resource names plugin.


andreamlin:
Do tests cover these changes?

lukesneeringer:
> Do tests cover these changes?

The smoke tests should serve as proof (which means something is wrong right now).

alexander-fenster:
@lukesneeringer do we still need this?

lukesneeringer:
I think we can assume this is dead.

PR #606, opened 2019-04-22
Update gapic-generator hash

andreamlin:
:D

I'll wait for the github release v0.16.27 to be tagged before merging this


andreamlin:
PTAL

alexander-fenster:
I'm starting to hate all these hashes! Debugging things will be a nightmare :)

andreamlin:
@alexander-fenster what would be better? like listing exactly which gapic generator commits are included?

alexander-fenster:
@andreamlin Until we deprecate Artman, I believe we will benefit if we mention gapic-generator changes in the release notes, and update Artman version accordingly (not just patch release, but minor or major as well).

In general, the main problem is that Artman patch version bumps have nothing to do with the functionality being added/removed. For the users gapic-generator is a hidden part of the Artman black box, so whenever any new feature makes its way to gapic-generator, we _probably_ should bump minor (for back-compatible changes) or major (for breaking changes) Artman version, with some description in the release notes, even though the Artman code itself does not change. This will probably break the 'letter' of the semantic versioning, but will surely support its 'spirit'.

Example: adding `x-goog-request-params` headers, in my opinion, was worth bumping a minor, not just a patch.

andreamlin:
i will have to fix the bugs

PR #616, opened 2019-05-07
Allow searching outside google folder for a gapic directory

Katee:


test/tasks/test_python_grpc.py
```diff
@@ -22,33 +22,33 @@
...
 
-def test_get_subdir_path():
+def test_get_gapic_subdir_path():
     task = python_grpc_tasks.PythonMoveProtosTask()
     with mock.patch.object(os, 'walk') as walk:
         walk.return_value = (
-            ('foo', ['bar'], []),
-            ('foo/bar', ['proto'], []),
-            ('foo/bar/proto', [], ['file_pb2.py']),
+            ('project', ['v1'], []),
```
alexander-fenster: Can we add a test that covers new functionality (the `docs` and `tests` directory that contains `gapic` subdirectory and it not returned by `get_gapic_subdir_path`?

alexander-fenster:
Thanks @Katee! I'm OK with this change, let's just have the new logic covered by a unit test.

test/tasks/test_python_grpc.py
Katee: It doesn't matter if `docs` or `tests` include a gapic folder, `os.walk` will never be run on them. I've added a mock of `os.listdir` showing this.

PR #618, opened 2019-05-16
Update gapic-generator hash

alexander-fenster:
Including some fixes for Node.js: now depending on the gax version 1.1.0 and requiring Node.js v8.10+. Also, no dependency on `lodash.merge` anymore.


alexander-fenster:
@andreamlin OK, then just let me know when gapic-generator is in a good shape and I'll update the hash. Not merging this one for now. Thank you!

alexander-fenster:
@andreamlin done! But it should wait for #619.

PR #623, opened 2019-05-17
Release v0.20.0

alexander-fenster:
Releasing new version (with incompatible changes for Node.js). I'll merge this when googleapis hash is updated (#622).


alexander-fenster:
Let's wait for decision in #621 and then release.

PR #624, opened 2019-05-22
Java: Generate samples in samples-google-cloud-api-version directory

hzyi-google:
gapic-generator puts samples in `src/samples`. This PR basically does:

```shell
mv src/samples ../samples-google-cloud-{api}-{version}
```

from the directory of gapic-google-cloud-{api}-{version}.


artman/config/converter.py
```diff
@@ -80,7 +80,7 @@ def _calculate_rel_gapic_output_dir(language, api_name, api_version):
     ouput dir becomes configurable. This logic doesn't work for non-cloud.
     """
     if language == 'java':
-        return 'java/gapic-google-cloud-%s-%s' % (api_name, api_version)
+        return 'java'
```
andreamlin: Can you comment here that gapic-generator will take care of the API-specific directory name?

hzyi-google: @andreamlin thanks for reviewing. I actually have some more changes coming up the these two PRs so please hold off :)

hzyi-google:
Based on the offline discussion with @vam-google, it doesn't seem to be a good idea to let gapic-generator decide the package name. So instead, gapic-generator will just output gapics in `gapic/src/main/java` and samples in `samples/src/main/java`. Artman will rename `gapic` and `samples` to appropriate package name.

I've updated the title and description too. 

@andreamlin @vam-google PTAL

alexander-fenster:
@hzyi-google Is this PR still actual or can we close it? Doing some housekeeping, you know.

hzyi-google:
@alexander-fenster I still need this (it went stale because Vadym was out for a while). Now that @vam-google is back I need to get the PR up to date for Vadym to take a look.

hzyi-google:
@vam-google PTAL

hzyi-google:
No longer needed.

PR #636, opened 2019-06-07
Add proto_package option

andreamlin:
Pass a `--package` option to gapic-generate via a `--proto_package` option in artman.

If a gapic config is not given to gapic-generator, then we need a `--package` option in gapic-generator to narrow down the list of protofiles from which to generate a client.


andreamlin:
PTAL

test/config/testdata/expected_library_config.yaml
```diff
@@ -0,0 +1,25 @@
...
+  proto_package: google.example.library.v1
+  service_yaml: /tmp/input/google/example/library/library.yaml
+  src_proto_path:
+  - /tmp/input/google/example/library/v1
+  test_proto_deps:
+  - name: google-iam-v1
+    proto_path: google/iam/v1
+java:
+  gapic_code_dir: /tmp/output/java/gapic-google-cloud-library-v1
+  release_level: ga
```
hzyi-google: newline at eof?

test/tasks/test_gapic.py
```diff
@@ -158,7 +158,8 @@ def test_execute(self, exec_command, check_output):
             service_yaml='/path/to/service.yaml',
             toolkit_path='/path/to/toolkit',
             aspect='ALL',
-            generator_args='--extra_args'
+            generator_args='--extra_args',
+            proto_package=''
```
hzyi-google: Would it be possible to add `proto_package` to `generator_args`?

andreamlin: What goes inside of generator args? I figure this would best be put in the artman file, because (1) that is where the gapic_yaml option is, and (2) it is an unchanging config value that we shouldn't require the user to write out on the command-line on every invocation.

test/config/testdata/expected_library_config.yaml
andreamlin: Done!

andreamlin:
fixed things
PTAL

test/tasks/test_gapic.py
hzyi-google: Oh I see.

hzyi-google:
LGTM

PR #637, opened 2019-06-07
Specify different versions of protoc for each language

andreamlin:
Each language has its own version of protoc that is used to do all the protoc stuff.

There have been at least two incidents where different languages had a protobuf version ceiling that was lower than the latest released version; the most recent incident is how google-cloud-java is pinned to the same version of protobuf that grpc-java uses, which is 3.7.1, even though the other languages use 3.8.0.


andreamlin:
PTAL

Dockerfile
```diff
@@ -64,21 +76,31 @@ RUN apt-get update \
     zlib1g \
   && rm -rf /var/lib/apt/lists/*
 
-# Install protoc 3.8.0.
-RUN mkdir -p /usr/src/protoc/ \
-  && curl --location https://github.com/google/protobuf/releases/download/v3.8.0/protoc-3.8.0-linux-x86_64.zip > /usr/src/protoc/protoc.zip \
-  && cd /usr/src/protoc/ \
+# Install all versions of protoc needed by all the languages.
+
+# Install protoc ${PROTOBUF_PYTHON_VERSION} for Python and other languages that use this version.
```
hzyi-google: Using Python for other languages seems a bit less straightforward.

Would it be possible first figure out all the versions of protoc needed and then download and link them?

andreamlin: Ok! Moved the logic into a bash script `install_protoc.sh`.

andreamlin:
PTAL

vam-google:
@andreamlin Why was this needed? Why didn't we just update java's dependency to be 3.8.0 as well? Making proto version per language complicates the system even more. 

What did you mean by java being inned to the same version as grpc-java? How is "pinn" implemented?

alexander-fenster:
Yeah, I also kind of dislike the idea of parsing the shell script :)  #649 is just one result of this (and it also highlights the sad fact that `--local` is not covered by any tests, makes me 😢 )  I guess we'll need to refactor it somehow :)

PR #640, opened 2019-06-12
Release v0.25.0

andreamlin:
Contains a new feature to pass in a `proto_package` parameter via the artman YAML, to the gapic-generator's `--package` option. This allows artman to generate clients without a GAPIC yaml.


andreamlin:
PTAL

PR #641, opened 2019-06-14
Release v0.26.0

andreamlin:
- New feature to use protoc 3.7.1 for Java client generation, and ability to set language-specific versions of protoc


andreamlin:
PTAL

PR #650, opened 2019-06-26
fix --local runs

### LINKED ISSUE #649: --local runs are broken
After #637, the `--local` runs don't work since the logic of `install_protoc.sh` does not work outside of the carefully crafted Docker image. We need to use `install_protoc.sh` in Docker only, and if it does not exist, not fail but revert to using the `protoc` installed in the system.

alexander-fenster:
Fixes #649. The `protoc` versions PR #637 broke `--local` runs. My understanding is that the PR approach is just wrong and it should be reverted and/or changed, but if we don't revert it right now, let's at least make local runs work by not failing because of the missing script and having reasonable default value for `protoc` command (which is, obviously, just `protoc` without any specific version).


alexander-fenster:
@vam-google If you have a merge conflict while reverting, feel free to add me and I'll fix it. Or, feel free to revert both since my change does not make any sense without #637.

PR #655, opened 2019-06-27
use protoc v3.9.0-rc-1 for PHP

alexander-fenster:
Ads ask to try using `protoc` v3.9.0-rc-1 for PHP (download location is [here](https://github.com/protocolbuffers/protobuf/releases/download/v3.9.0-rc1/protoc-3.9.0-rc-1-linux-x86_64.zip)). Internal bug ID: 131809012.

@vam-google: it might be difficult to stop using this protoc override feature :(  We need to think what other options we have, since this case shows that using one version across all languages might be really needed (PHP needs to use RC1 ASAP which we likely cannot use across all languages).


vam-google:
@alexander-fenster For ads it is doable, becaus they will have their own WORKSPACE file, where they can specify different version of protoc. I.e. it is ok to have different protobuf version per "org" (Ads or Cloud), but not different protobuf version per language.

vam-google:
@alexander-fenster Notice what is happening here: Ads request affects cloud as well (i.e. if Ads is Ok using 3.9.0-rc1 for PHP (i.e. non-stable version) it does not mean that Google Cloud is Ok doing same).

dwsupplee:
@vam-google @alexander-fenster 

> Ads request affects cloud as well (i.e. if Ads is Ok using 3.9.0-rc1 for PHP (i.e. non-stable version) it does not mean that Google Cloud is Ok doing same).

This is spot on, and unfortunately we weren't okay with using a release candidate. We've got a backlog of PRs which we can't progress on now - I originally came here to see if we could pin to v3.8 at generation time but see we have a challenge given the ads team and their requirements. Are there any recommended solutions?

PR #662, opened 2019-07-02
Add SampleGen smoke tests to CircleCI

### LINKED ISSUE #654: Add sample generation to golden test (to catch samplegen regressions)
We just accidentally pushed a release which broke samplegen – no sample files being output:
 - https://github.com/googleapis/artman/issues/653

To catch this, let's update the golden library so it has at least 1 code sample.

Here are some of the bits I've found out so far (really @alexander-fenster did) ~

 - We'll add the names of the generated sample files here:

https://github.com/googleapis/artman/blob/22e143fcdb01e18caa6fc58e1ae3d5d3e2f90dce/test/golden/library_example.golden#L10

 - We'll need to add `--generator-args="--dev_samples"` to these gapic invocations:

https://github.com/googleapis/artman/blob/22e143fcdb01e18caa6fc58e1ae3d5d3e2f90dce/.circleci/config.yml#L42

> _I think? If these are the golden?_

Here's where the golden gapics are invoked so maybe I'm wrong about it being part of the smoketest above :)

https://github.com/googleapis/artman/blob/22e143fcdb01e18caa6fc58e1ae3d5d3e2f90dce/test/golden/artman_golden_test.py#L51

So, yeah. Need to dive in more, find the source for the golden library (proto + gapic config) and add samples.

@googleapis/samplegen 
@alexander-fenster

hzyi-google:
Fix #654

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,14 +90,14 @@ jobs:
             EXIT_STATUS=0
             docker run -it --rm -d --name samplegen --workdir /tmp -e RUNNING_IN_ARTMAN_DOCKER=True artman
             docker exec samplegen mkdir /tmp/language-python
-            docker exec samplegen artman --local --output-dir /tmp/language-python --config=google/cloud/language/artman_language_v1.yaml --root-dir=/googleapis --generator-args='--dev_samples' generate python_gapic
-            docker exec samplegen find /tmp/language-python/python/language-v1/samples/v1 -mindepth 1 || EXIT_STATUS=$?
+            docker exec samplegen artman --local --output-dir /tmp/language-python --config=google/cloud/language/artman_language_v1.yaml --root-dir=/googleapis --generator-args='--dev_samples' generate python_gapic || EXIT_STATUS=$?
+            docker exec samplegen find /tmp/language-python/python/language-v1/samples/v1 -mindepth 1 | read || EXIT_STATUS=$?
             docker exec samplegen mkdir /tmp/talent-php
-            docker exec samplegen artman --local --output-dir /tmp/talent-php --config=google/cloud/talent/artman_talent_v4beta1.yaml --root-dir=/googleapis --generator-args='--dev_samples' generate php_gapic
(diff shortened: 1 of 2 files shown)


hzyi-google:
@alexander-fenster Do you happen to know how to make the samplegen_smoke_test required?

alexander-fenster:
@lukesneeringer can do it (you need to be a repo admin to change this setting).

hzyi-google:
@alexander-fenster Thanks!
Luke is out this week. Can you take a look at the changes first? I'll bug him next week when he gets back to get this test required.

.circleci/config.yml
```diff
@@ -74,6 +74,33 @@ jobs:
...
+      - *load_build_and_save_artman_image
+      - *save_artman_cache
+      - run:
+          name: Check samples are generated for selected APIs
+          command: |
+            EXIT_STATUS=0
+            docker run -it --rm -d --name samplegen --workdir /tmp -e RUNNING_IN_ARTMAN_DOCKER=True artman
+            docker exec samplegen mkdir /tmp/language-python
+            docker exec samplegen artman --local --output-dir /tmp/language-python --config=google/cloud/language/artman_language_v1.yaml --root-dir=/googleapis --generator-args='--dev_samples' generate python_gapic
+            docker exec samplegen find /tmp/language-python/python/language-v1/samples/v1 -mindepth 1 || EXIT_STATUS=$?
```
alexander-fenster: Here you only check that the `samples/v1` folder is generated, but you don't actually check that any files are there. It's better than no check at all, but I would probably check that the number of Python files inside is greater than zero (same for PHP and other language).

alexander-fenster:
```
fenster-macbookpro:tmp fenster$ mkdir empty
fenster-macbookpro:tmp fenster$ ls empty
fenster-macbookpro:tmp fenster$ find empty | read || echo bad
fenster-macbookpro:tmp fenster$ find empty | read
fenster-macbookpro:tmp fenster$ echo $?
0
```

hzyi-google:
It seems with `-mindepth 1` it will work (otherwise `.` is always an entry there)

```
tmp $ mkdir empty
tmp $ find empty -mindepth 1 | read || echo bad
bad
```

alexander-fenster:
Oh! That makes sense.

PR #669, opened 2019-07-18
DiscogapicConfigGen: Use a temporary output directory

hzyi-google:
`DiscogapicConfigGen` generates gapic yaml from discovery docs, and therefore the flag `gapic_yaml` is used as output instead of input (which is the usual case for codegen).

When trying to refresh gapic config for compute artman creates this default `artman-genfiles` in my local environment while does not drop any output there. Therefore it seems clearer to just use a temporary directory to hold the generated gapic config before moving to the right place.

Also, if user specifies the `output_dir` flag when invoking `artman generate discogapic_config`, throw a command line warning saying that the flag is ignored, and output is saved in the directory given by  `gapic_yaml`.


hzyi-google:
Hmmmm @alexander-fenster How should I fix the tests if this case...?

hzyi-google:
PTAL

artman/cli/main.py
```diff
@@ -273,6 +274,10 @@ def normalize_flags(flags, user_config):
     elif artifact_type == Artifact.DISCOGAPIC_CONFIG:
         pipeline_name = 'DiscoGapicConfigPipeline'
         pipeline_args['discovery_doc'] = artifact_config.discovery_doc
+        if flags.output_dir != os.path.abspath(DEFAULT_OUTPUT_DIR):
```
alexander-fenster: what if it's the same but just not the absolute path? maybe use `os.path.abspath(flags.output_dir)` as well to compare apples to apples?

hzyi-google: `output_dir` is already converted to absolute path here: https://github.com/googleapis/artman/pull/669/files/1a72f234a170a66b904b932488b8c801b5c5adad#diff-371ed57b95bd30de3b3331f9cc2ff446L217

But yeah, I see the point here. Adding the `abspath` here will make it less confusing.

PR #672, opened 2019-07-24
Update gapic-generator hash

hzyi-google:
Adding the following features from gapic-generator:
- Java: Remove unsupported retry samples
- Java: Add `@generated` to generated package-info files
- Node.js: drop dependency on through2


hzyi-google:
@googleapis/artman-reviewer

PR #677, opened 2019-07-31
Add a `gapic-samples` field to artman config and pass sample configs to gapic-generator

hzyi-google:


hzyi-google:
cc/ @googleapis/samplegen

PR #678, opened 2019-07-31
node: compile protos into JSON

alexander-fenster:
This PR adds some logic for artman post-processing that compiles all the proto files into JSON for `nodejs_gapic` tasks. The resulting JSON file (that will appear in all client libraries) is not yet used but very soon will be!

Also, upgrading Node.js inside artman Docker image to v10 (it's now v4 but it was not used anywhere before this PR :) ), and installing `google-gax` globally (to have the `compileProtos` script).


Dockerfile
```diff
@@ -62,8 +58,21 @@ RUN apt-get update \
...
 
+# Download and unpack Node.js v10
+RUN mkdir -p /opt \
+  && curl -L https://nodejs.org/dist/v10.16.0/node-v10.16.0-linux-x64.tar.xz -o /opt/node-v10.16.0-linux-x64.tar.xz \
+  && tar -C /opt -xJf /opt/node-v10.16.0-linux-x64.tar.xz \
+  && rm -f /opt/node-v10.16.0-linux-x64.tar.xz
+ENV PATH /opt/node-v10.16.0-linux-x64/bin:$PATH
+
+# Install google-gax for Node.js
+RUN npm install -g google-gax@^1.2.1
```
JustinBeckwith: Oh bother.  What's going to keep that version up to date?

Dockerfile
```diff
@@ -62,8 +58,21 @@ RUN apt-get update \
     libunwind8 \
     libuuid1 \
     zlib1g \
+    xz-utils \
   && rm -rf /var/lib/apt/lists/*
 
+# Download and unpack Node.js v10
+RUN mkdir -p /opt \
+  && curl -L https://nodejs.org/dist/v10.16.0/node-v10.16.0-linux-x64.tar.xz -o /opt/node-v10.16.0-linux-x64.tar.xz \
```
JustinBeckwith: Little worried about how we keep the version of nodejs you're running here up to date.

alexander-fenster: So, first of all, it does not make it any worse because previously it was Node 4 from Ubuntu :)  Then, this version of Node.js is only used for compiling protos. I suggest we leave both versions (Node.js and google-gax) hardcoded here for now (same way as we do for other stuff in this `Dockerfile` such as `protoc` version, etc.) and update as needed. 

Actually, we'll get rid of artman soon (I hope so) so who cares?

Dockerfile
alexander-fenster: Well... semver `^1.2.1` will allow everything up to `2.0`, and it actually prevents me from possible breaking changes in `google-gax`. It's pretty much the same way as we do in `package.json`, isn't it?

PR #734, opened 2019-09-17
Update gapic-generator hash

hzyi-google:
to include the following features from gapic-generator:
* Remove rogue entrySet call when iterating over maps in C# samples
* C# samples: fix incorrect bracing style and and a newline at EOF
* SampleGen: import OperationFuture and XxxMetadata for LRO samples

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM ubuntu:16.04
 
 # Release parameters
 ENV GOOGLEAPIS_HASH 47bd0c2ba33c28dd624a65dad382e02bb61d1618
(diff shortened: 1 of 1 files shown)


alexander-fenster:
Let's commit it now so we have more CI tests passed, but we'll likely need https://github.com/googleapis/gapic-generator/pull/2979 soon (so one more hash update will be needed).

hzyi-google:
@alexander-fenster I want to cut a release right away - the last change listed above will unblock us releasing first Java samples :)

I can do another release when the resource name fixes are in.

PR #736, opened 2019-09-19
Fix Resource Names annotations

vam-google:
Fixes common resource names case.
Depends on:
- https://github.com/googleapis/gapic-generator/pull/2979
- https://github.com/googleapis/protoc-java-resource-names-plugin/pull/59


artman/tasks/protoc_tasks.py
```diff
@@ -120,33 +141,50 @@ class ProtoCodeGenTask(ProtocCodeGenTaskBase):
...
         return self._execute_proto_codegen(
             language, src_proto_path, import_proto_path, pkg_dir,
             api_name, api_version, organization_name, toolkit_path,
-            gapic_yaml, gen_proto=True,
+            gapic_yaml, root_dir, gen_proto=True,
             final_src_proto_path=final_src_proto_path,
             final_import_proto_path=final_import_proto_path,
             excluded_proto_path=excluded_proto_path)
 
+class ResourceNameGenTask(ProtocCodeGenTaskBase):
```
hzyi-google: Just curious: what do you have to add this task? Shouldn't artman already be able to invoke the resource name plugin and generate resource name classes?

vam-google: I had to split generation of resource names and generation of protobuf stubs. Since now plugin does not use insertion poiints, nothing prevents us from doing it. I had to do it to prevent generation of proto stubs for the common resource names messages - since to declare annotation we need a message, `common_resources.proto` declares those "fake" messages which exist solely to declare resource annotation in them.

hzyi-google: That makes sense.

PR #741, opened 2019-09-24
release 0.37.1 with whitelisted test

alexander-fenster:
Releasing the patch release with just one change: a test with a wrong PHP configuration is whitelisted so that it does not cause the whole CI failure.


lukesneeringer:
Why does this require a release?

alexander-fenster:
I was under impression that googleapis CI uses the latest released artman, not the master artman - that's why releasing a patch.

PR #751, opened 2019-10-21
Update gapic generator hash

hzyi-google:
include the following fix from gapic generator:
- SampleGen: update Node.js const name for library package (#2969)


hzyi-google:
cc/ @beccasaurus

PR #752, opened 2019-10-22
Release 0.40.2

hzyi-google:
To include the following feature from gapic-generator:
- SampleGen: update Node.js const name for library package (#2969)


hzyi-google:
cc/ @beccasaurus

PR #755, opened 2019-10-28
config: add grpc_service_config attr to config

noahdietz:
* updates config.proto
* updates config_pb2.py
* updates config loading/converting code
* updates testdata
* updates gapic_tasks to pass the `--grpc_service_config` generate arg using config value


artman/tasks/gapic_tasks.py
```diff
@@ -103,7 +103,7 @@ class GapicCodeGenTask(task_base.TaskBase):
     def execute(self, language, toolkit_path, descriptor_set, service_yaml,
```
alexander-fenster: A function with 14 parameters makes me cry. What can go wrong here :)

(obviously needs a refactor, but I'm guessing it will never happen)

alexander-fenster:
One important thing that might be useful here is to change a smoke test to pass this config to artman if it exists in the API folder. Do you think it makes sense to add?

PR #771, opened 2019-12-11
update googleapis hash

alexander-fenster:
It turns out we might need to update internal googleapis copy (inside the Docker image) since we updated `google/cloud/common_resources.proto`. I'm really unsure how exactly it is used, but we got a bug report that this file is missing, and this is the first place I thought of. Let's try updating this hash.


hzyi-google:
LGTM

PR #784, opened 2020-01-13
run protoc in one invocation (and stop testing Python 2.7)

alexander-fenster:
In this PR:

- make sure we pass all the protos to the same `protoc` invocation for all languages except Go. For Go, keep doing what we are doing: split list of protos into several invocation based on `go_package` option.
- added tests that cover this functionality.
- when adding those tests, figured out that the code being changed does not work on Python 2.7 at all. Since it's dead and we use 3.x inside Docker anyway, let's stop testing on 2.7.
- also, made sure our tests work with the latest `nox` version.

All in the same PR - sorry for that!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -157,23 +157,10 @@ jobs:
       - *checkout_artman
       - run:
           name: Install nox.
-          command: pip install --upgrade nox-automation==0.19.1
+          command: pip install --upgrade nox-automation
       - run:
           name: Build the docs.
-          command: nox -e docs
-    working_directory: /usr/src/artman/
-
-  unit-python2.7:
-    docker:
-      - image: python:2.7
-    steps:
-      - checkout
-      - run:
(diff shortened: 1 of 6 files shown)


alexander-fenster:
@pierrickvoulet - FYI, this is for you :)

noahdietz:
> For Go, keep doing what we are doing: split list of protos into several invocation based on go_package option.

I'm not clear on _why_ this needs to happen. Can you elaborate?

alexander-fenster:
@noahdietz No idea. Was introduced by @pongad in https://github.com/googleapis/artman/pull/539 (and previously in https://github.com/googleapis/artman/pull/536), but I don't have any details and I'm just keeping the Go behavior as is.

noahdietz:
> @noahdietz No idea. Was introduced by @pongad in #539 (and previously in #536), but I don't have any details and I'm just keeping the Go behavior as is.

Ah OK. Thanks.

PR #786, opened 2020-01-15
make python gRPC task parse import public in protos

alexander-fenster:
`import public` in protos [is a thing](https://developers.google.com/protocol-buffers/docs/proto#importing-definitions), and it seems like in Python there is a code that changes

```proto
import "google/cloud/securitycenter/v1/run_asset_discovery_response.proto";
```

to

```proto
import "google/cloud/securitycenter_v1/proto/run_asset_discovery_response.proto";
```

but this logic fails to update `import public` statement, which causes `File not found` error. This is a blind attempt to fix it (I don't see an easy way to verify that this fix works).


hzyi-google:
Did you (is it possible) to verify it locally?

alexander-fenster:
@hzyi-google I wasn't able yet :)  What I will do is I will just run the whole smoke test here, I bet it will be easier than making it run locally.

alexander-fenster:
@hzyi-google Running tests [here](https://circleci.com/gh/googleapis/artman/13133?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link), we'll see if securitycenter passes or not.

alexander-fenster:
@hzyi-google It worked!

image

image

(this one failure is expected, it's the same as in master and unrelated; securitycenter worked though!)

PR #825, opened 2020-04-10
Update gapic-generator hash and Java resource name plugin

hzyi-google:
gapic-generator:
- feat: add the last segment when generating names for singleton resoures
- fix: add null check and better error message when referenced resources are not found
- feat: Support extra plugin_args for php bazel rules rules (#3165)
- feat: support '*' in resource definition (#3163)
- fix: align text in license (#3158)
- fix(python): add PROJECT_ID for python-system-test (#3161)
- fix(PYTHON): update nox file and modernize to new interface (#3156)

resource name plugin
- feat: add the last segment when generating names for singleton resources
- feat: support * annotation in resource def (#84) 
- feat: trace up only one level when calculating parent types with singleton resources
- fix: stop generating `parseList` and `toStringList` if a multi-pattern resource class has subclasses

also whitelist Asset to unblock releasing artman.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM ubuntu:16.04
 
 # Release parameters
 ENV GOOGLEAPIS_HASH e47fdd266542386e5e7346697f90476e96dc7ee8
-ENV GAPIC_GENERATOR_HASH ef5d22b920d73f6b996e012d7235313c4ebc4428
+ENV GAPIC_GENERATOR_HASH 8d53d31494bebc1e1da4ed0d1352f1f35441d439
 # Define version number below. The ARTMAN_VERSION line is parsed by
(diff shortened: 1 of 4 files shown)


hzyi-google:
_Looking at golden test..._

alexander-fenster:
Golden test is basically just a list of filenames that gets compared to the output. Seems like Python has changed `noxfile.py` to `nox.py`? It's in https://github.com/googleapis/gapic-generator/pull/3156.

hzyi-google:
@alexander-fenster done!

alexander-fenster:
Isn't it cool to release artman in your last day in the team @hzyi-google ? :)

hzyi-google:
Hahaha and I even looked into the golden test directory for the first time :)

